PR #28, opened 2020-02-12
V8 lockers

darind:
In this PR we are introducing `v8::Locker` objects to the codebase that allow accessing the isolate from multiple threads. This removes the need from marshalling javascript callbacks to the main thread.

### DIFF
--- a/NativeScript/NativeScript.mm
+++ b/NativeScript/NativeScript.mm
@@ -32,12 +32,15 @@ static std::shared_ptr<Runtime> runtime_;
     runtime_ = std::make_shared<Runtime>();
 
     std::chrono::high_resolution_clock::time_point t1 = std::chrono::high_resolution_clock::now();
-    runtime_->Init();
+    Isolate* isolate = runtime_->CreateIsolate();
+    runtime_->Init(isolate);
     std::chrono::high_resolution_clock::time_point t2 = std::chrono::high_resolution_clock::now();
     auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(t2 - t1).count();
     printf("Runtime initialization took %llims\n", duration);
 
     if (config.IsDebug) {
+        Isolate::Scope isolate_scope(isolate);
+        HandleScope handle_scope(isolate);
         v8_inspector::JsV8InspectorClient* inspectorClient = new v8_inspector::JsV8InspectorClient(runtime_.get());
         inspectorClient->init();
         inspectorClient->registerModules();
--- a/NativeScript/inspector/JsV8InspectorClient.mm
+++ b/NativeScript/inspector/JsV8InspectorClient.mm
@@ -124,10 +124,8 @@ void JsV8InspectorClient::init() {
     }
 
     Isolate* isolate = runtime_->GetIsolate();
-    Isolate::Scope isolate_scope(isolate);
-    HandleScope handle_scope(isolate);
 
-    Local<Context> context = isolate->GetCurrentContext();
+    Local<Context> context = isolate->GetEnteredOrMicrotaskContext();
 
     inspector_ = V8Inspector::create(isolate, this);
 
@@ -150,6 +148,7 @@ void JsV8InspectorClient::createInspectorSession() {
 
 void JsV8InspectorClient::disconnect() {
     Isolate* isolate = runtime_->GetIsolate();
+    v8::Locker locker(isolate);
     Isolate::Scope isolate_scope(isolate);
     HandleScope handle_scope(isolate);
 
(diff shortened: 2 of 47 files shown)


farfromrefug:
Very interesting. That could mean js scripts could run on a multi thread environment?

NathanaelA:
@darind - Any chance this can be applied to the Android runtime also?

darind:
@farfromrefug, yes this PR enables scripts to run in a multi-threaded environment.

For example the following script will print _false_:

```javascript
const queue = NSOperationQueue.alloc().init();
queue.addOperationWithBlock(() => {
    // running on a background thread
    console.log(NSThread.currentThread.isMainThread); 
});
```

Of course this doesn't mean that scripts can run in parallel. And that's the big difference between this javascript and the corresponding native code:

```objective-c
[[[NSOperationQueue alloc] init] addOperationWithBlock:^{
    NSLog(@"%s", [[NSThread currentThread] isMainThread] ? "true" : "false");
}];
```

We have already discussed it but let's repeat it again so that there are no confusions or false expectations:

> V8 isolates are not reentrant by design.

You can never run javascript on the same isolate from multiple threads **in parallel**. In the javascript sample above, if the main thread is busy executing some script, the background thread will simply wait until it finishes before it is able to acquire the lock and run. And again while the background thread is running your callback, the main thread will not be able to execute any javascript (it can execute some other stuff, but not javascript because it will not be able to acquire the lock).

So the bottomline from this is that if you really need parallel javascript execution, the proper way to achieve this is to use Workers or separate isolates.

darind:
@NathanaelA, yes this can be applied to the Android runtime as well.

The problem is that this will be a catastrophic breaking change rendering the modules and existing plugins useless because (at least in Android) authors have always relied that the code they write in their javascript callbacks will run on the main thread.

The reason we are doing it for the V8 iOS beta runtime is simply to make it behave like the existing iOS runtime.

farfromrefug:
@darind I know it does not mean parallel and it is not what i want.
What i am looking for is the current IOS runtime behavior: JS callback are run from their native calling thread. Which is awesome! For example it means you can have realtime camera processing.


Yes i want exactly that for android and we already discussed that many times. And yes that change would mean all plugins would have to change their code to make the actual "runOnMainThread" when necessary.
But i do think it is a must have and it can actually mean an increase in performance because of the removal of `runOnUiThread` which is currently done in the android runtime for all JS calls.

BTW it would be an imrpovement for map plugins for example. Mapbox, CartoSDK, ... have callback which are actually not called on main thread. And it is intended so. In {N} those callbacks (JS side) are done on the UI thread which can slow down the UI.
Even if it "pauses" the JS thread because JS can't run in parallel this would be a huge improvement to have and would allow apps which are not possible today.

NathanaelA:
@darind - 
Can this patch be applied with a switch to optionally enable it on Android for testing through the rest of 6.x so that maybe we could get plugins ready and then it could be the default in NS 7?   

This should fix the issue Android OpenGL issues...  Which to my knowledge is one of the only places that NS can't actually do from JS.

darind:
@farfromrefug, agree with you that this can be beneficial to the Android runtime as well, but we do not have the resources to change all existing plugins. And even if we can do it for our plugins, we cannot force third party authors to do it. I am not sure that you realising the impact of such change on the ecosystem. It is not only making the change physically in the code of the plugins but also testing it. Also plugins can be combined making the job of testing even harder.

@NathanaelA, the change would be wayyyy too substantial to be a candidate for a feature flag.

farfromrefug:
@darind i get what you mean. Now could we run some tests to actually see how it goes? I mean how hard it is to implement just runtime wise? Do you think we could have a branch to test the change? And maybe it could even be a option like "multiThread" which would be false by default.

I would give a lot to try/have this :D

darind:
@farfromrefug, yes we can have a branch with the new functionality.

NathanaelA:
@darind - Hmm, yeah I realize the major change to all the plugins; I've had to do changes to the plugins in every major version of NS already anyways.   :-)   

Do you think this is something that the team would be willing to push towards in NS 7?   I don't know what the team has on the plate for NS7; and I know this requires a lot of work and testing.   But   this could be a major enhancement on the Android side, being able to stay in the thread we are in -- I've had to do some interesting work arounds when I ran into callbacks that expected it to be in the original thread.    

Maybe this "feature" branch could be like the v8-beta runtime on ios, a v8-locker runtime for android that gets automatically updates so that people like me and Martin, can test and start moving things over to it.    We can start moving plugins we test and fix into a v8-locker version tag.

darind:
Yes this is a feature that we can consider for a possible NS7 release.

You can track the progress of the Android runtime update here: https://github.com/NativeScript/android-runtime/pull/1587

darind:
@farfromrefug, @NathanaelA : an early preview of the functionality you've asked for is available in the `darind/v8-locker` branch.

Corresponding PR: https://github.com/NativeScript/android-runtime/pull/1587

Any feedback is welcome.

PR #65, opened 2020-10-01
fix: xcode 12 builds

rigor789:
Not sure if that refactor actually solves the issue: https://stackoverflow.com/questions/27256277/storing-stdto-stringx-c-str-in-member-variable-produces-garbage

It was a compile-time error with the prior way, this gets the build working, but not sure if the underlying issue (the reason for the compile error) is actually fixed by the change.


adrian-branescu:
That std::string has to be allocated on the heap, otherwise you will end up holding a dangling pointer.
When the scope of that loop will end the underlying buffer of the std::string (that char * you are using further) will get deallocated.

NathanaelA:
@adrian-branescu -

Thanks for the assist, definitely need more skilled C++ developers.

```
        auto nameStr = std::to_string(i);
        const char* name = nameStr.c_str();
        errors->SetName(name);
```
> That char* has to be allocated on the heap, otherwise you will end up holding a dangling pointer.
> When the scope of that loop will end the underlying buffer of the std::string (that char * you are using further) will get deallocated.

Unless I misunderstand you; That is correct, the value that we are creating is copied into its own local variable by the `errors->setName(name)` so after that point the `name`  and `nameStr` variable are actually no longer needed in this loop instance and they should be eliminated after the loop and function are complete and won't matter if they point to something invalid past the point of `setName`...

Or did I misunderstand what you were saying.   Because I will admit I haven't done as much C++ work recently, so I am a bit rusty.  :grinning:

adrian-branescu:
@NathanaelA 
I think that it depends of the implementation of that [setName](https://github.com/NativeScript/ns-v8ios-runtime/blob/master/NativeScript/inspector/third_party/inspector_protocol/crdtp/error_support.h#L36) method.

If it performs a strcpy or memcpy then it's fine, but judging by its signature I'm not sure it does that.

String literals reside in .rodata segment of the executable, so they are usually referenced throughout code as _const char *_ without performing extra copies on heap. Idk if the guys from V8 took any safety measure or not, probably it's best to check their code.

NathanaelA:
@adrian-branescu - You are correct; I went back to the v8 code and it is just a pointer assignment.    I had assumed (bad thing.  LOL) that the prior person who wrote the had checked if it was a copy, because of the code that I was fixing was written as if it was a copy.      Guess their code had two bugs. :)     I'll just do a malloc and memcopy so that we can save the value...

Again THANK you for the assist!

PR #66, opened 2020-10-10
chore: 7.0.2-RC.1 changes

NathanaelA:
This has the full change set to build under XCode 12.01 with LLVM 8.01

### DIFF
--- a/NativeScript/inspector/utils.h
+++ b/NativeScript/inspector/utils.h
@@ -23,7 +23,8 @@ static std::unique_ptr<protocol::Array<T>> fromValue(protocol::Value* value, pro
     std::unique_ptr<protocol::Array<T>> result(new protocol::Array<T>());
     errors->Push();
     for (size_t i = 0; i < array->size(); ++i) {
-        const char* name = std::to_string(i).c_str();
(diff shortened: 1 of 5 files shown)


NativeScript/inspector/utils.h
```diff
@@ -22,8 +24,15 @@ static std::unique_ptr<protocol::Array<T>> fromValue(protocol::Value* value, pro
...
     errors->Push();
+    
+    // This (log10) will return the max number of digits(-1) in the length of the array that we have to worry about, +1 for the null
+    int sizeArray = log10(array->size()) + 2;
+    
     for (size_t i = 0; i < array->size(); ++i) {
-        const char* name = std::to_string(i).c_str();
+	// Convert array index to a string value
+	char* name = new char[sizeArray];
+	sprintf(name, "%lu", i);
```
NathanaelA, line 34: @adrian-branescu - What do you think about this solution;
Line 29, we figure out the max number of characters for the array size + 1 for null
Line 33, we create the value on the heap
Line 34, we use sprintf to convert the number to a string

The other method just was a lot more code for no pursoe, long to string, string -> cstr, alloc, then copy

Issue #67, opened 2020-10-10
Appears to be an issue with some classes on iOS 10


NathanaelA:
Basically, the first time the UILayoutGuide.alloc().init() is everything is created correctly, and the object works as expected.
The second time thru the routine, the UILayoutGuide.alloc() works; but the prototype chain is now missing the majority of the functions one of which is the `init` function, so then the application crashes with init not being a function.

- Appears to be directly related to the .init() call
- Can duplicate on a iOS 10 Emulator w/ XCode 12
- See https://github.com/NativeScript/NativeScript/pull/8954 for PR used to fix the issue.

Prototype Chain on first `alloc()` (120 items):
```
-> bottomAnchor
-> centerXAnchor
-> centerYAnchor
-> hasAmbiguousLayout
-> heightAnchor
-> identifier
-> layoutFrame
-> leadingAnchor
-> leftAnchor
-> owningView
-> rightAnchor
-> topAnchor
-> trailingAnchor
-> widthAnchor
-> constraintsAffectingLayoutForAxis
-> encodeWithCoder
-> initWithCoder
-> toString
-> accessibilityActivationPoint
-> accessibilityCustomActions
-> accessibilityCustomRotors
-> accessibilityElements
-> accessibilityElementsHidden
-> accessibilityFrame
-> accessibilityHint
-> accessibilityLabel
-> accessibilityLanguage
-> accessibilityNavigationStyle
-> accessibilityPath
-> accessibilityTraits
-> accessibilityValue
-> accessibilityViewIsModal
-> autoContentAccessingProxy
-> classForCoder
-> classForKeyedArchiver
-> isAccessibilityElement
-> observationInfo
-> shouldGroupAccessibilityChildren
-> accessibilityActivate
-> accessibilityAssistiveTechnologyFocusedIdentifiers
-> accessibilityDecrement
-> accessibilityElementAtIndex
-> accessibilityElementCount
-> accessibilityElementDidBecomeFocused
-> accessibilityElementDidLoseFocus
-> accessibilityElementIsFocused
-> accessibilityIncrement
-> accessibilityPerformEscape
-> accessibilityPerformMagicTap
-> accessibilityScroll
-> addObserverForKeyPathOptionsContext
-> awakeAfterUsingCoder
-> awakeFromNib
-> copy
-> dealloc
-> dictionaryWithValuesForKeys
-> didChangeValueForKey
-> didChangeValueForKeyWithSetMutationUsingObjects
[...]

PR #70, opened 2020-10-16
fix: Do not update in-place object prototypes (#67)

darind:
Related to #67 

This PR addresses an issue where we replace the existing object prototype with additional protocols returned from pseudo types. Instead of modifying the prototype chain, we build an entirely new chain containing the additional protocols.

### DIFF
--- a/NativeScript/runtime/ArgConverter.h
+++ b/NativeScript/runtime/ArgConverter.h
@@ -31,8 +31,8 @@ class ArgConverter {
 public:
     static void Init(v8::Local<v8::Context> context, v8::GenericNamedPropertyGetterCallback structPropertyGetter, v8::GenericNamedPropertySetterCallback structPropertySetter);
     static v8::Local<v8::Value> Invoke(v8::Local<v8::Context> context, Class klass, v8::Local<v8::Object> receiver, V8Args& args, const MethodMeta* meta, bool isMethodCallback);
-    static v8::Local<v8::Value> ConvertArgument(v8::Local<v8::Context> context, BaseDataWrapper* wrapper, bool skipGCRegistration = false);
-    static v8::Local<v8::Value> CreateJsWrapper(v8::Local<v8::Context> context, BaseDataWrapper* wrapper, v8::Local<v8::Object> receiver, bool skipGCRegistration = false);
+    static v8::Local<v8::Value> ConvertArgument(v8::Local<v8::Context> context, BaseDataWrapper* wrapper, bool skipGCRegistration = false, const std::vector<std::string>& additionalProtocols = std::vector<std::string>());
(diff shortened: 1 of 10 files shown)


NathanaelA:
@darind - This PR did fix the original issue; I ran tests and the framework no longer crashes and the prototype chain looked correct (afaik) on the second+ runs.     WooHoo!

However, This PR seems to have either broken a couple things (or exposed them); 
1st, on the iOS 10 simulator this PR now fails `transitionCoordinator.animateAlongsideTransitionCompletion` automated tests (And other tests like wise, but this seems to be the easiest that always fails on my iOS 10 simulator, every time)...
```
CONSOLE ERROR: Test: --- [ROOT-VIEWS-CSS-CLASSES.test_modal_root_view_modal_css_class] FAILED: transitionCoordinator.animateAlongsideTransitionCompletion is not a function, Stack: TypeError: transitionCoordinator.animateAlongsideTransitionCompletion is not a function
at Page._showNativeModalView (file: src/Users/nathanael/Source/core/NativeScript/packages/core/ui/core/view/index.ts:497:25)
```

2nd, on the iOS 14 simulator this crashes right away with:
```
CONSOLE INFO: Test: START ROOT-VIEWS-CSS-CLASSES TESTS.
====== Assertion failed ======
Native stack trace:
1          0x10661a3d9 tns::Assert(bool, v8::Isolate*) + 119
2          0x106615528 tns::ClassBuilder::ExposeProperties(v8::Isolate*, objc_class*, std::__1::vector<tns::PropertyMeta const*, std::__1::allocator<tns::PropertyMeta const*> >, v8::Local<v8::Object>, v8::Local<v8::Value>, v8::Local<v8::Value>)::$_1::__invoke(ffi_cif*, void*, void**, void*) + 344
3          0x1066d22b6 ffi_closure_unix64_inner + 662
4          0x1066d353a ffi_closure_unix64 + 72
5       0x7fff23ca4d51 -[UIStatusBarManager _updateStyleForWindow:animationParameters:] + 66
```

---
Step to duplicate:
- create a dist release of this PR as you need the .tgz file
- checkout the `github/nativescript` repo
- cd nativescript
- npm run setup
- cd nativescript/apps/automated
- tns platform add ios --frameworkPath=<TGZ> file you created
- npm i
(to select a single test group & make the tests much quicker; you can manually edit src/test-runner.ts on line 26; set `singleTest` to `ROOT-VIEWS-CSS-CLASSES`
image

- tns run ios --env.testing

darind:
@NathanaelA, very good catch! Thanks for verifying. I pushed an additional commit which should address this issue. Can you check it now?

By the way, I saw that there are 4 failing tests on iOS 14 and a crash on iOS 10.3 with the official runtime from the `master` branch when running all test suites. I see the exact same behaviour with the runtime from this PR.

NathanaelA:
Yes, I will double check, thanks for the quick fixes!     And yes their are apparently some crashes unrelated to this PR, I plan on creating some issues/fixes today on them once I trace down if it is runtime or core modules related....

NathanaelA:
Awesome PR - This fixes the issue and the later issues I brought up.    Thank you!

PR #73, opened 2020-10-20
feat: Faster loading of JS files

NathanaelA:
This PR creates a custom version of the code that will loads the JS modules (i.e. require(module)).   By not creating extra buffers and extra memory usage, this speeds up the loading of JS files by ~16%.

### DIFF
--- a/NativeScript/runtime/Helpers.h
+++ b/NativeScript/runtime/Helpers.h
@@ -23,6 +23,7 @@ bool ToBool(const v8::Local<v8::Value>& value);
 std::vector<uint16_t> ToVector(const std::string& value);
 
 bool Exists(const char* fullPath);
+v8::Local<v8::String> ReadModule(v8::Isolate* isolate, const std::string &filePath);
 const char* ReadText(const std::string& filePath, long& length, bool& isNew);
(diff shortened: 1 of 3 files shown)


darind:
This is great optimization. Just note that with the single webpack bundle (leading to a single require) and code cache (in release mode), the effect of this optimization is mitigated. Actually this function will never be used in the hot path of an application run in release mode.

NathanaelA:
>This is great optimization. Just note that with the single webpack bundle (leading to a single require) and code cache (in release mode), the effect of this optimization is mitigated. Actually this function will never be used in the hot path of an application run in release mode. Nevertheless, this optimization will improve load times during development and is worth including.

Unless I'm mistaken even with Code-Cache it still has to load the two massive Vendor.js and Bundle.js files doesn't it.      Guess I need to double check with Code-Cache on...  :+1:    (Currently dealing with a app that has a > 10m files in Dev mode)

darind:
Yes, you are right. Vendor and Bundle are still loaded from disk, even with code-cache on. But it's still worth checking the performance gain with an app built in Release and running on an actual device.

Issue #51, opened 2020-06-29
Sporadic crash on app exit - iOS 13.5.1


NathanWalker:
In a current app on TestFlight, some testers are encountering this occasionally when the app exits (full crashlog attached):

```
Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Triggered by Thread:  0

Thread 0 name:
Thread 0 Crashed:
0   libsystem_kernel.dylib        	0x00000001a4456df0 __pthread_kill + 8
1   libsystem_pthread.dylib       	0x00000001a4376930 pthread_kill + 228 (pthread.c:1458)
2   libsystem_c.dylib             	0x00000001a4304c24 __abort + 116 (abort.c:147)
3   libsystem_c.dylib             	0x00000001a4304bb0 abort + 116 (abort.c:118)
4   libc++abi.dylib               	0x00000001a4429f40 abort_message + 132 (abort_message.cpp:76)
5   libc++abi.dylib               	0x00000001a441b8b0 demangling_terminate_handler() + 284 (cxa_default_handlers.cpp:60)
6   libobjc.A.dylib               	0x00000001a4384f1c _objc_terminate() + 148 (objc-exception.mm:704)
7   TNSCrashlyticsLogger          	0x0000000108849cb8 CLSTerminateHandler() + 348
8   libc++abi.dylib               	0x00000001a44293cc std::__terminate(void (*)()) + 20 (cxa_handlers.cpp:59)
9   libc++abi.dylib               	0x00000001a4429358 std::terminate() + 44 (cxa_handlers.cpp:88)
10  NativeScript                  	0x0000000105f0ed60 std::__1::shared_ptr<v8::Persistent<v8::Value, v8::NonCopyablePersistentTraits<v8::Value> > >::~shared_ptr() + 64
11  libsystem_c.dylib             	0x00000001a42e5164 __cxa_finalize_ranges + 416 (atexit.c:284)
12  libsystem_c.dylib             	0x00000001a42e54a0 exit + 28 (exit.c:81)
13  UIKitCore                     	0x00000001a87beb88 -[UIApplication _terminateWithStatus:] + 508 (UIApplication.m:6735)
[...]

PR #74, opened 2020-10-21
fix: crash at shutdown

NathanaelA:
Removes the unmatched ->Exit() that was left over from prior refactor and is causing random crashes on exit of application.
Closes: #51


NathanaelA:
We tested this PR in Production (7.0.4); the Dispose will throw now frequently on swipe away as the Isolate is still in a "Enter" state.  The original code was correct, and so this will be revered out of 7.0.5

PR #90, opened 2020-12-15
fix: DictionaryAdapter missing handle scopes and locks

delaneyb:
Fixes https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/1728, which seems to have been due to thread switching in the callback at [FIRFunctions.m#L187](https://github.com/firebase/firebase-ios-sdk/blob/62735099d51778e43f15adc8f0057b7c8d72e17b/Functions/FirebaseFunctions/FIRFunctions.m#L187). This issue will affect any native library functions called with a JavaScript object which are then wrapped in a DictionaryAdapter, and then attempt to access or enumerate the DictionaryAdapters contents from another thread.

I have not worked in Objective C nor with V8 internals before. This pull request needs a careful review and testing. Thank you @darind for pointing me in the right direction.

### DIFF
--- a/NativeScript/runtime/DictionaryAdapter.mm
+++ b/NativeScript/runtime/DictionaryAdapter.mm
@@ -34,6 +34,10 @@ using namespace tns;
 
 - (id)nextObject {
     Isolate* isolate = self->isolate_;
+    v8::Locker locker(isolate);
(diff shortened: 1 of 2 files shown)


darind (approved):
Looks good to me 👍

rigor789 (approved):

PR #93, opened 2020-12-29
fix: PromiseProxy returns function regardless of underlying property type

delaneyb:
The current wrapper causes accesses to any property of a promise object to return a function, even when the actual value on the target proxy is not a function.

I don't understand why it's only for properties **other than** .then/.catch that the proxy trap interferes and makes sure the function executes on the same thread as the Promise was created, but I'll leave that for someone else to look into.


rigor789 (approved):

NathanWalker:
Thank you @delaneyb this was merged in with your other PR here https://github.com/NativeScript/ns-v8ios-runtime/pull/90/files - and will be released shortly.

PR #94, opened 2020-12-30
Support for Apple Silicon

darind:
- Added arm64 architecture to support new Apple M1 chips
- Added arm64 build for the metadata generator
- Added LLVM 8.0 as git submodule dependency


rigor789:
@andresilva-cc this is not the only change required, so it takes a while to build every piece of the tooling for arm64. Rosetta2 should be working however, until everything is updated.

andresilva-cc:
@rigor789 I know, but according to [this](https://gist.github.com/shirakaba/87aabee6e2ec70e298e3d7c42f701675#gistcomment-3579260), that would make things a lot easier.

neil-119:
Also looking forward to this.

rigor789:
Closing, since this has been merged as part of #108.

PR #108, opened 2021-03-18
feat: apple silicon support

rigor789:
This is #94 with some additional changes in the build scripts to simplify & drop usage of `lipo` (unsupported and not recommended by Apple).

~Not ready to merge just yet, some cleanup required with the default project structure around TNSWidgets~ Was a DerivedData cache issue.


hraynaud:
@rigor789 is this fully released yet?

davorpeic:
@hraynaud Don't think so as building from terminal fails, I manage to run the simple template app only from xcode when modifying architecture, so I guess silicons are not yet fully supported

Issue #100, opened 2021-01-21
Promise.resolve and Promise.reject causes memory leak on iOS.


schnapzz:
Simply referencing another issue posted on the Nativescript repo but it seem to be an issue with the v8 ios runtime so I'll post it here as it seem a very important issue to be fixed.

https://github.com/NativeScript/NativeScript/issues/9151.

delaneyb:
@darind This part of PromiseProxy seems to be be the culprit: https://github.com/NativeScript/ns-v8ios-runtime/blob/92195e8f7c02aa50a97d686757e0b04070c78ef7/NativeScript/runtime/PromiseProxy.cpp#L18-L26

Replacing with
```
let promise = new target(origFunc)
```
or commenting out the script in PromiseProxy entirely fixed the leak for me

Any immediate ideas?

rigor789:
The reason this code exists is to ensure the promise callbacks are always executed on the same thread where they are constructed. Though I don't quite understand the inner `Proxy` by just reading the code, will need to run & see what it does.

delaneyb:
Not exactly clear what situation it becomes useful as long as V8 locking is used properly, and I'm not sure it works fully as intended using try-catch blocks in async functions (.catch() will get invoked on the same thread but I think the mechanics of try catch in terms of execution moving to the catch block are different)

Anyway, I dug further, this triggers the same memory leak:
```javascript
let count = 0
function cb () {
    ++count
}

setInterval(() => {
    let runloop = CFRunLoopGetCurrent();
    for (var i = 0; i < 100; i++) {
        CFRunLoopPerformBlock(runloop, kCFRunLoopDefaultMode, cb);
        CFRunLoopWakeUp(runloop);
    }
    console.log(count);
}, 50);
```

It's the cb being passed to CFRunLoopPerformBlock and how it's processed from there I believe... My guess would be that it's a challenge for the runtime to know when CFRunLoopPerformBlock is done invoking `cb`, and hence when to clean up anything allocated...

Is there a free some where to go with this, which is getting invoked as part of translating the cb so it can be invoked by ObjC: https://github.com/NativeScript/ns-v8ios-runtime/blob/8409f9af84de73e2cd00cbeebce3ba4b53431a6e/NativeScript/runtime/Interop.mm#L70

delaneyb:
[...]

PR #110, opened 2021-04-08
fix: Blocks memory leak (#100)

darind:
Related to #100

### DIFF
--- a/NativeScript/runtime/Interop.h
+++ b/NativeScript/runtime/Interop.h
@@ -126,6 +126,7 @@ public:
     static id ToObject(v8::Local<v8::Context> context, v8::Local<v8::Value> arg);
     static v8::Local<v8::Value> GetPrimitiveReturnType(v8::Local<v8::Context> context, BinaryTypeEncodingType type, BaseCall* call);
 private:
+    static std::pair<IMP, ffi_closure*> CreateMethodInternal(const uint8_t initialParamIndex, const uint8_t argsCount, const TypeEncoding* typeEncoding, FFIMethodCallback callback, void* userData);
     static CFTypeRef CreateBlock(const uint8_t initialParamIndex, const uint8_t argsCount, const TypeEncoding* typeEncoding, FFIMethodCallback callback, void* userData);
     template <typename T>
     static void SetStructValue(v8::Local<v8::Value> value, void* destBuffer, ptrdiff_t position);
@@ -186,7 +187,8 @@ private:
         const void* invoke;
         JSBlockDescriptor* descriptor;
         void* userData;
-
+        ffi_closure* ffiClosure;
+        
         static JSBlockDescriptor kJSBlockDescriptor;
     } JSBlock;
 };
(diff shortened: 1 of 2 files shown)


brysem:
Hi @darind,

Thank you for your effort! I've tested this PR in one of the larger apps I've made. The issue from #100 still seems to be present.

I tested it by using the app as normal and monitoring memory usage.

I've also tested it with a button that executes the following piece of code. This piece of code dramatically increases memory usage in V8. In JSC it does not.
```js
var dispatch = function() {
    setTimeout(function() {
    for(var i=0; i < 100000; i++) {
        var p = new Promise((resolve, reject) => { resolve('') })
    }
    dispatch()
    }, 50)
}
dispatch();
```

darind:
@brysem, by creating 100k promises every 50ms you are putting too much pressure on the JS engine. Here's the profiler graph when I reduce it to 5 seconds:

image

Make sure you leave some CPU cycles for the GC to complete.

NathanaelA:
@darind - Did you try this runtime with a fresh NS app on a real device; I am doing a:

(Using latest TNS 8 cli)
- ns create demo --js        
- cd demo
- ns platform add ios --frameworkPath <this_compliled_version>
- ns run ios

And I am seeing on a real ios 14.2 (iPhone7) device:

image

I'm not sure if the memory is being freed to early and so we are getting corrupted data back or what exactly...

darind:
@NathanaelA, no I haven’t tried this on a real device in a fresh NS app. Does this crash occur in the Simulator?

But looking at the error, this looks like some issue with passing C strings as arguments to functions: https://github.com/NativeScript/ns-v8ios-runtime/blob/master/NativeScript/runtime/Interop.mm#L150

Are you sure that the crash doesn’t occur with a build from the master branch? Also is the use of the NSGetSizeAndAlignment function something new to the NS 8 core modules? Has this function been used before?

NathanaelA:
@darind - 
> Does this crash occur in the Simulator?

Unknown, I can try a simulator if you need; I just tend to use a real device for most my testing.

> Are you sure that the crash doesn’t occur with a build from the master branch? 

Yes, to be honest, I thought it was my build initially, but I finally went back to `master` and built it and it worked, then I built another version with the other two outstanding PR's and both those worked.   However, the two different builds that I tried  that included this specific PR caused the above error.

> Also is the use of the NSGetSizeAndAlignment function something new to the NS 8 core modules? Has this function been used before?

That command does not appear to exist in the source code directly, I believe it is a side effect of the transition invocation.   However the area that the call stack points to hasn't been changed in a long time (I believe this code has remained unchanged and existed before NS6, but for sure it has existed for all of NS7)

NathanaelA:
@darind - Hold off, I'm now really confused.  (BTW: The broken builds also don't work on simulators.)    Going to re-checkout master, and re-do all the builds and see what happens.   

I know the second build I created yesterday was only this PR, but I just created another new build using just this PR and it is working.      So, give me a bit of time to figure out what is different between the multiple branches I've created...

NathanaelA:
@darind - I apologize; it isn't your PR.  Sorry to waste your time.    It is actually one of the other PR's, I suspect some how I got build artifacts from my first build where I included all three PR's...   Today after making sure everything is cleaned each time I narrowed down which PR properly)

NathanWalker:
This pr is rolled up in https://github.com/NativeScript/ns-v8ios-runtime/pull/111 and in `rc` for further validations.

PR #111, opened 2021-04-09
fix: memory leaks & WeakRefs

rigor789:
WIP


brysem:
Would it be possible to get a RC release for testing?

NathanaelA:
@rigor789 - Something in this PR is breaking NS apps.     However the other two PR's you merged into this one seems to be working fine with all the tests...
  
(Using latest TNS 8 cli)
- ns create demo --js        
- cd demo
- ns platform add ios --frameworkPath <this_compliled_version>
- ns run ios

And I am seeing on a real ios 14.2 (iPhone7) device (& 14.0 simulators):

image

rigor789:
It's the `strdup` change, it's excluded from the alpha build, just haven't found a good solution for it's leak yet...

NathanaelA:
@rigor789 -

> It's the `strdup` change, it's excluded from the alpha build, just haven't found a good solution for it's leak yet...

So the alpha has all these changes except the strdup change, I assume you are talking about this strdup change?
https://github.com/NativeScript/ns-v8ios-runtime/pull/111/commits/0f96dc4a2a6d014335528bc2e91b193b7696348c#diff-bec6e3d6b4fa0eacfb7726d6e5915c9eecda4e319bc61f71a0527e740eb197ddR160

NathanaelA:
@brysem,
> Would it be possible to get a RC release for testing?

The `tns platform add ios@alpha` according to Igor above has all these fixes except for a single broken one he manually backed out.   However, in my tests of the @alpha engine it has some interesting side effects.   I haven't traced them all down yet, but WeakRefs seem to take a LOT longer to be collected (not sure if they always are (???), but in my tests it does appear that after about 7 seconds they will be collected, but they don't appear to be collected from manually calling a GC event.       So we would love to know if you are seeing any major memory issues in your app when using the Alpha version.    

If you would like you can also try this version (a version I built for a different nStudio client):
https://nativescript.rocks/clients/nativescript-ios-8.0.1-rc1.tgz -- it only contains the two other PR's (which are also in this PR) and doesn't include any of the other changes in this PR, so it will still be leaking some memory that Igor fixed, but should fix the majority of the big leaks...   This rc1 we will try and get pushed to NPM soon as it at least appears to fix several issues but causes no new issues, where the alpha is failing about 10 tests (all GC / weakref related).

rigor789:
> So the alpha has all these changes except the strdup change, I assume you are talking about this strdup change?
0f96dc4#diff-bec6e3d6b4fa0eacfb7726d6e5915c9eecda4e319bc61f71a0527e740eb197ddR160
> 
> Yeah, that strdup needs to be present (or something that makes a copy of the data, as once it leaves the scope it looks like the value is freed).

That one is a tricky one - the data in some cases does persist fine without the `strdup` call, and is released once no longer needed, but sometimes that data is released too early. Having that `strdup` there does leak all the memory though, since it's never released afterwards. This is a big issue for apps that rely heavily on SQLite, since as far as I understand - all the queries are going through that exact branch - for most apps, it's probably not a huge deal if it only leaks a few times (still bad, but tolerable).

But yes, the latest alpha has that one change reverted before build - because it does cause the failure you are seeing in some cases (garbage passed instead of the right value, because it's freed too early).

@darind suggested we try making the arguments into `Persistent` values and then add them to a global cache where we can get notified when the argument is eligible for releasing, but I couldn't test that yet (won't be able to for the next 2 weeks).

The fastest way to debug/step through and reach that branch is to call
```js
NSMethodSignature.signatureWithObjCTypes('v@:c');
```

if you want to experiment with that idea.

---

> WeakRefs seem to take a LOT longer to be collected

v8's WeakRefs make no guarantees when they are actually released or GC'd - with this PR, we are switching to the built-in WeakRef implementation, so that change is expected...

NathanaelA:
Thanks for the insight Igor;   I was thinking making it a `shared_ptr` might also work.   As once the shared_ptr is no longer referenced it would call `free` automatically.  

Something like `shared_ptr<unsigned char> ptr(strdup(...), free);`   But that code is fairly complicated with lots of indirection's, so the getter code would have to be changed to use a .get() to get the pointer out of it for usage.     But it is something I might try when I have some free time...   :grinning:

brysem:
> If you would like you can also try this version (a version I built for a different nStudio client)
Thanks. I will give it a try today.

As already mentioned by Igor the change on https://github.com/NativeScript/ns-v8ios-runtime/commit/0f96dc4a2a6d014335528bc2e91b193b7696348c#diff-bec6e3d6b4fa0eacfb7726d6e5915c9eecda4e319bc61f71a0527e740eb197ddR160 does indeed create a lot of weird issues for the nativescript-sqlite package. It also broke something in frame transitions from the core (if my memory serves me right).

PR #114, opened 2021-04-12
DictionaryAdapter Hanging References

brysem:
DictionaryAdapter is leaking memory when objects are deallocated.

This PR serves to clean up some hanging references. In my test case this reduced the total leaks caused by DictionaryAdapter by 36%.


brysem:
@cla-bot check

PR #118, opened 2021-06-02
fix: boolean to NSNumber marshalling

triniwiz:
fixes: https://github.com/NativeScript/NativeScript/issues/9429
fixes: https://github.com/NativeScript/nativescript-ui-charts/issues/17


cjohn001:
@triniwiz: Thanks for looking into it!

PR #119, opened 2021-07-06
chore: 8.1 prep

NathanWalker:


NathanWalker:
to whom it may concern, @edusperoni and I have been validating this against some projects and found some additional issues and crashes. We were stepping through some of it today and started a branch to begin clearing a few which is considered experimental while we investigate:
https://github.com/NativeScript/ns-v8ios-runtime/tree/fix/memleak-stabilization

NathanWalker:
Obsolete by https://github.com/NativeScript/ns-v8ios-runtime/pull/123/files

PR #141, opened 2022-01-25
fix: prevent crashes during onuncaughterror

### LINKED ISSUE #126: discardUncaughtJsExceptions: true not working with new runtime
Hello together,
I just came over an issue were a plugin throws an exception which lets my app crash. With the JSC runtime I can set 
discardUncaughtJsExceptions: true  
in nativescript.config.ts

and than I only get an error message as follows on the console:
NativeScript discarding uncaught JS exception!

instead of the app crashing. Seems like the new runtime does not implement this behavior. It simply always crashes.
Would be great if you could implement this option for the new runtime as well

Archez:
This PR aims to address some unwanted crashes that would happen during the `OnUncaughtError` handler.

The first relates to how the `AppPackageJson` file was being read to fetch the `discardUncaughtJsExceptions` flag. The NSJSONSerialization would produce a NSDictionary that would get autoreleased. On subsequent calls to `OnUncaughtError`, the `AppPackageJson` pointer would be cleared causing a bad access.

To address this, I alloc and init a new NSDictionary from the serialization so that this dictionary will not get released between calls to `OnUncaughtError`. I'm not super familiar with Objective-C or how this runtime is setup, so I don't know if there is a point where we _should_ manually release this dictionary. Please let me know if this should be handled.

Second, I ran into an unhelpful stack trace when trying to test the error events I setup and eventually realized that my event handlers were also causing a javascript exception, but since we were already in the `OnUncaughtError` method, the `tns::Assert` would be false and quit the app. I addressed this by using `TryCatch` and logging the caught error so developers have visibility on if their error event handlers are also causing an error.

This PR will most likely close #126 and possibly is related to some crashes other users have reported.

### DIFF
--- a/NativeScript/runtime/NativeScriptException.mm
+++ b/NativeScript/runtime/NativeScriptException.mm
@@ -43,7 +43,12 @@ void NativeScriptException::OnUncaughtError(Local<Message> message, Local<Value>
         Local<Object> thiz = Object::New(isolate);
         Local<Value> args[] = { error };
         Local<Value> result;
+        TryCatch tc(isolate);
         success = errorHandlerFunc->Call(context, thiz, 1, args).ToLocal(&result);
+        if (tc.HasCaught()) {
+            tns::LogError(isolate, tc);
+        }
+
         tns::Assert(success, isolate);
     }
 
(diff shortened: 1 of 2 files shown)


NathanWalker:
We will release this in the next 8.2.x alpha shortly thank you for this.

NathanWalker:
In [8.2.0-alpha.6](https://www.npmjs.com/package/@nativescript/ios/v/8.2.0-alpha.6) now.

PR #142, opened 2022-02-03
Improvements for crash report

dgmachado:


NativeScript/runtime/Helpers.mm
```diff
@@ -691,3 +691,13 @@
     Log(@"%s", stack.c_str());
     assert(false);
 }
+
+void tns::StopExecutionAndLogStackTraceIfInDebug(v8::Isolate* isolate) {
```
edusperoni, line 695: rename to `StopExecutionAndLogStackTrace` or remove

NativeScript/runtime/Helpers.mm
```diff
@@ -691,3 +691,13 @@
...
+    Assert(false, isolate);
+}
+
+void tns::LogDebugMessage(std::string message) {
+    if (RuntimeConfig.IsDebug) {
+        Log(@"%s", message.c_str());
+    }
+}
```
edusperoni, line 703: remove and use `Log` directly

NativeScript/runtime/MetadataBuilder.mm
```diff
@@ -810,6 +810,14 @@
     Class klass = objc_getClass(containingClass.c_str());
     // TODO: Find out if the isMethodCallback property can be determined based on a UITableViewController.prototype.viewDidLoad.call(this) or super.viewDidLoad() call
 
+    NSString* message = [NSString stringWithFormat:@"MetadataBuilder::InvokeMethod class {%s}, selector {%s}, isInitializer {%s}, type {%s}, lib {%s}",
+                         containingClass.c_str(),
+                         meta->selectorAsString(),
+                         meta->isInitializer() ? "true":"false",
+                         meta->typeName(),
+                         meta->topLevelModule()->getName()];
+    tns::LogDebugMessage(message.UTF8String);
```
edusperoni, line 819:
```
if(debug) {
   NSString* message = [NSString stringWithFormat:@"MetadataBuilder::InvokeMethod class {%s}, selector {%s}, isInitializer {%s}, type {%s}, lib {%s}",
                         containingClass.c_str(),
                         meta->selectorAsString(),
                         meta->isInitializer() ? "true":"false",
                         meta->typeName(),
                         meta->topLevelModule()->getName()];
    tns::Log(message.UTF8String);
}
```

if you don't check for debug first, you'll waste processing on building the message and never show it

NativeScript/runtime/Helpers.mm
edusperoni: to clarify. This function will induce people to make the mistake of building a big heavy string and discarding it here.

NativeScript/runtime/Helpers.mm
dgmachado: Done

NativeScript/runtime/MetadataBuilder.mm
dgmachado: Done

PR #155, opened 2022-03-17
fix(inspector): ensure socket message is copied and stored

edusperoni:
also adds a nice semaphore so we convert a busy wait into a more CPU friendly one


edusperoni:
@NathanWalker it seems this does solve the iOS crash when debugging with the devtools. Still not sure why this is even needed as it's just adding some "inefficient" code to ensure we're copying strings and retaining them.

To expand further in the issue, the message is received correctly, gets into the `onFrontendMessageReceived` correctly and with the correct string but somewhere between there and the `pop_back()` the string has invalid data, so you either get garbage in it and it fails a conversion or `pop_back()` crashes saying the internal std::string storage (cstring) isn't even allocated to begin with. My suspiscion is that somehow the `std::string message` was being deleted (maybe due to the fact that it comes from `dispatch_data_t`) before the block executed, so garbage data was being pushed back.

Not confident in my explanation. The issue was too rare for a proper debugging/memory inspection, but this change seems to fix it.

PR #177, opened 2022-07-25
feat: split long logs into multiple instead of truncating

rigor789:
NSLog has a 1024 character limit, and anything beyond is truncated with `<...>`. This PR aims to split longer logs into multiple `NSLog` calls when necessary.

This is already possible on Android by setting the `maxLogcatObjectSize` boolean under the android key in the config.

The goal is to add a similar config option for iOS as well.

### DIFF
--- a/NativeScript/runtime/Console.cpp
+++ b/NativeScript/runtime/Console.cpp
@@ -34,6 +34,22 @@ void Console::Init(Local<Context> context) {
     }
 }
 
+void Console::SplitAndLogInChunks(std::string message) {
+    auto messageLength = message.length();
+    int maxStringLength = 1000; // technically 1024, but let's have some room :)
+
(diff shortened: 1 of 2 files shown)


farfromrefug:
@rigor789 will the result of the splitted log be on the same line in the console? Like let s say it is a JSON.stringify which is splitted. Will you be able to copy it, paste it as a correct JSON string?

farfromrefug:
@rigor789 i am intersted in testing this. Any chance to have an alpha runtime with this?

edusperoni:
@farfromrefug once the build finishes you can get the .tgz from the artifacts here: https://github.com/NativeScript/ios/actions/runs/5456662035

rigor789:
Thanks @edusperoni 👍

PR #188, opened 2022-12-12
fix: throw NSException on main thread

farfromrefug:
This helps catchers like `NSSetUncaughtExceptionHandler` get the correct info from the `NSException`


farfromrefug:
Thanks @NathanWalker will help a lot with sentry

PR #195, opened 2023-03-01
Expose runScriptString public method

tdermendjiev:
This PR enables NativeScript embedders to run arbitrary js strings with the NS runtime.


edusperoni:
@tdermendjiev can this be closed in favor of https://github.com/NativeScript/ios/pull/231?

PR #197, opened 2023-03-21
build_nativescript.sh quality of life improvements

caitp:
This adds some command line processing for build_nativescript.sh, so that messing with the scripts isn't strictly needed each time it's invoked during testing.

This was helpful while struggling with the build process, and made it easier to skip the catalyst builds for the final stretch


NathanWalker:
obsolete as already merged to `dev`

PR #218, opened 2023-06-19
feat: v8 version 11

NathanWalker:


caitp:
The current design has CPPGC owning the v8::Platform object, so this required undoing the previous ownership change in which Platform had shared ownership. However, the Cppgc platform itself is shared, and will retain the v8 platform until all references to it are disposed.

However, it's possible there are still tests that fail due to trying to keep the v8 platform alive by grabbing Runtime::GetPlatform(), retaining the platform will need to happen via Runtime::GetCppgcPlatform() instead now.

caitp:
The test failures are caused by the worker tests, as we know workers are not currently adapted to the cppgc landsacpe in this branch as of yet. With `runWorkerTests()` commented out in runAllTests(), everything appears to be passing.

PR #226, opened 2023-08-24
Re-add cleanup code from DisposeValue

caitp:
This re-introduces a bunch of Caches cleanup from the ObjectManager, which may improve the situation with leaks to some degree. More work is (probably) needed here.

To be honest, I've only verified that these are passing the in-tree tests, have not tried it against a large application that would see problems from the eternal objects held by the Caches.

### DIFF
--- a/NativeScript/runtime/DataWrapper.mm
+++ b/NativeScript/runtime/DataWrapper.mm
@@ -1,0 +1,81 @@
+#include "DataWrapper.h"
+
+#include "Caches.h"
+#include "FFICall.h"
+#include "ObjectManager.h"
+
+namespace tns {
+
+StructWrapper::~StructWrapper() {
+    parent_.Reset();
+    if (data_) {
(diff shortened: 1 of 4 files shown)


NativeScript/runtime/DataWrapper.mm
```diff
@@ -0,0 +1,81 @@
...
+#include "Caches.h"
+#include "FFICall.h"
+#include "ObjectManager.h"
+
+namespace tns {
+
+StructWrapper::~StructWrapper() {
+    parent_.Reset();
+    if (data_) {
+        // v8::Isolate::GetCurrent() may not necessarily be the isolate this object lives in.
```
caitp, line 12: (but I sure hope it is.. I mean really, it should be.. if it's not, that seems like a v8-cppgc bug?)

NathanWalker:
This needs to be tested against larger projects - it appears to work against the ultimate tab app.

PR #231, opened 2023-10-19
feat: allow embedding into Swift host projects

tdermendjiev:
This PR includes methods required for running NS apps and js strings from a native app using the NS runtime:
- (void)runScriptString: (NSString*) script runLoop: (BOOL) runLoop;
- (void)restartWithConfig:(Config*)config;
- (void)shutdownRuntime;

A default metadata binary is added to the framework product. This contains metadata about all the Apple frameworks. 
TBD: Create a separate build target for the framework used for embedding.


NativeScript/NativeScript.mm
```diff
@@ -21,18 +21,68 @@ @implementation Config
...
+
+- (void)shutdownRuntime {
+    if (runtime_ != nullptr) {
+        Isolate* isolate = runtime_->GetIsolate();
+        {
+            v8::Locker l(isolate);
+            v8::Isolate::Scope isolate_scope(isolate);
+            v8::HandleScope handle_scope(isolate);
+            
+            isolate->Dispose();
```
edusperoni, line 67: this likely shouldn't be done here (the runtime destructor is responsible for deleting this isolate)

NativeScript/NativeScript.mm
```diff
@@ -59,25 +109,20 @@ - (instancetype)initWithConfig:(Config*)config {
         }
     }
     
-    return self;
```
edusperoni, line 62: should this still returns self?

NativeScript/NativeScript.mm
```diff
@@ -59,25 +109,20 @@ - (instancetype)initWithConfig:(Config*)config {
...
 }
 
-- (void)runMainApplication {
-    runtime_->RunMainScript();
-
-    CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0, true);
+- (instancetype)initWithConfig:(Config*)config {
+    if (self = [super init]) {
+        [self initializeWithConfig:config];
+    }
```
edusperoni, line 117: init will be called twice if this method is called (one here and another in initializeWithConfig)

v8ios.xcodeproj/project.pbxproj
NathanWalker, line 1: Are all these changes necessary in this pbxproj file @tdermendjiev ?

edusperoni:
can we just confirm this all works with existing projects? I'm not sure about the implications on the changes to modules in the pbxproj. But once that's confirmed I believe this can be merged

v8ios.xcodeproj/project.pbxproj
```diff
@@ -392,6 +395,7 @@
 			dstSubfolderSpec = 10;
 			files = (
 				C20525802577D6F900C12A5C /* NativeScript.framework in Embed Frameworks */,
+				2B5088A82BBEC1BC00F6EB68 /* TNSWidgets.xcframework in Embed Frameworks */,
```
edusperoni, line 398: is this addition necessary?

NativeScript/metadata-arm64.bin
edusperoni, line 1: how do we generate this again when the metadata is updated? Shouldn't this just be part of the build process?

PR #234, opened 2024-01-18
feat: URL & URLSearchParams

triniwiz:


edusperoni:
maybe this (and the android counterpart) can be written using the new node way of binding modules (so no changes needed in the Runtime.h/mm):

https://github.com/NativeScript/ios/pull/221/files

Also, I think we should just use "URL" instead of URLImpl for the property

triniwiz:
This seems way nicer

shirakaba:
See my [pre-review comments](https://github.com/NativeScript/android/pull/1801#issuecomment-1899507865) on the equivalent Android PR – I've got the same questions about this repo as well. And thanks again!

NativeScript/runtime/Runtime.mm
```diff
@@ -176,6 +178,80 @@ void DisposeIsolateWhenPossible(Isolate* isolate) {
...
     Console::Init(context);
     WeakRef::Init(context);
+    
+    
+    auto blob_methods = R"js(
+    const BLOB_STORE = new Map();
+    URL.createObjectURL = function (object, options = null) {
+        try {
+            if (object instanceof Blob || object instanceof File) {
+                const id = NSUUID.UUID().UUIDString;
```
edusperoni, line 188: @triniwiz you need to convert the UUID to lowercase. iOS generates UUID in uppercase but the standard is to use lower case:

> Are UUIDs Case-sensitive?
> No, UUIDs are written in base 16 which uses numbers 0-9 and characters a-f. There is no distinction between upper and lowercase letters. However, [RCF 4122 section 3](https://tools.ietf.org/html/rfc4122#section-3) requires that UUID generators output in lowercase and systems accept UUIDs in upper and lowercase.

NathanWalker (approved):

PR #248, opened 2024-04-15
fix: Xcode 15.3+ not setting TARGET_OS_IOS correctly

### LINKED ISSUE #245: Metadata filtering - module class ignored by the runtime
Hi,

I am bout trying to fixe a bug on `nativescript-image-cache-it` some class are ignored by the runtime with the metadata filtering ?
There are a lot of class included but the SDAnimatedImageView not, the problem that it is not excluded or something, it does not appear in the stderr  at all !
I've tried several runtime version and still the same ! but the plugin worked before !
this is the file :
```json
{
  "uses": [
    "SDWebImage.*:*", // I've added this line
    "SDWebImage*:*",
    "Metal.MTLDevice:*",
    "CoreImage.CIContext:*",
    "CoreImage.CIColor:*",
    "CoreImage.CIFilter:*",
    "CoreImage.CIImage:*",
    "CoreImage.CIVector:*",
    "Dispatch*:*",
    "Darwin*:*",
    "ObjectiveC*:*",
    "ObjectiveCRuntime*:*",
    "OpenGLES*:*",
    "nsswiftsupport:ImageCacheItUtils*"
  ]
}
```

Logikgate:
This stems from my work on a plugin that uses the SDWebImage package. Metadata for some classes were not getting generated when building for a physical device. 

This will probably also resolve #245 since they were also running into problems with the same SDWebImage package.

### DIFF
--- a/metadata-generator/build-step-metadata-generator.py
+++ b/metadata-generator/build-step-metadata-generator.py
@@ -152,10 +152,11 @@ def generate_metadata(arch):
                              deployment_target_flag_name + "=" + deployment_target])
     else:
       generator_call.extend(["-target", "{}-{}-{}{}".format(arch, llvm_target_triple_vendor, llvm_target_triple_os_version, llvm_target_triple_suffix)])
-      # since iPhoneOS 17.4 sdk TARGET_OS_IPHONE is not defined for non-simulator builds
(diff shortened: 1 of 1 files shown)


Logikgate:
@cla-bot check

rigor789 (approved):

Issue #249, opened 2024-04-17
Support for xrOS platform broke Apple Watch embedded apps


gabrielbiga:
After the merge of https://github.com/NativeScript/ios/pull/235, if the application has an Apple Watch embedded app, it will broke the compilation.

![image](https://github.com/NativeScript/ios/assets/1100522/75c2f7e9-6cf7-4ee9-8cab-97d2c3e86ab8)

Actually cleaning the `Swift-ObjC-Bridging-Header.h` file solves the issue. Its currently patched here because I don't know the correct way to fix it.

Platform: [v8.7.0](https://github.com/NativeScript/ios/releases/tag/v8.7.0)
Reverting to [v8.6.4](https://github.com/NativeScript/ios/releases/tag/v8.6.4) fix the issue.
Commit impacted: https://github.com/NativeScript/ios/commit/bb364f9558c336c43a9c43d3ded46ef1ad8e8bf3

PR #250, opened 2024-04-19
fix(ios): Fix broken build of WatchOS embedded apps

gabrielbiga:
This commit aims to fix this issue reported here: https://github.com/NativeScript/ios/issues/249

I don't know if this is the best solution, but as we don't need NS headers at WatchOS development, I think we can disable it for the target `TARGET_OS_WATCH`.


NathanWalker:
Hey thanks @gabrielbiga this seems like a reasonable fix - if this confirmed fixes on your side we could include this on a patch in next few days.

NathanWalker (approved):

gabrielbiga:
Hello @NathanWalker thanks for the review! The piece of code is tested and did solve the issue here. 👍🏻

NathanWalker (approved):

PR #265, opened 2024-12-17
feat: use monotonic time for performance object

edusperoni:
note: DO NOT SQUASH


NathanWalker (approved):

edusperoni:
We need to do the same on Android before 9.0 so both times are consistent

PR #276, opened 2025-07-16
feat: ES modules (ESM) support with conditional esm or commonjs consumption + better error handling

NathanWalker:
- Conditionally detects .mjs (esm) or .js bundled code
- Loads and executes ES modules
- When module or javascript errors occur, they are caught and proper file/line numbers are printed for better diagnosing

Backwards compatible.

## New `nativescript.config` options

1. `logScriptLoading: true`: Used to enable verbose terminal output on module resolution.
2. `showErrorDisplay: true`: Used to display an error screen on the device when errors are encountered. Off by default. Some may prefer to see in terminal only as some will still allow app to continue fine. Other teams may want priority put on all errors that occur so rendering an error screen the moment they happen forces developer to handle them.

### DIFF
--- a/NativeScript/NativeScript-Prefix.pch
+++ b/NativeScript/NativeScript-Prefix.pch
@@ -1,7 +1,7 @@
 #ifndef NativeScript_Prefix_pch
 #define NativeScript_Prefix_pch
 
-#define NATIVESCRIPT_VERSION "9.0.0"
+#define NATIVESCRIPT_VERSION "8.9.3"
 
 #ifdef DEBUG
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@nativescript/ios",
   "description": "NativeScript Runtime for iOS",
-  "version": "8.9.5",
+  "version": "9.0.0-alpha.8",
```
NathanWalker:
```suggestion
  "version": "8.9.5",
```

triniwiz (approved):

PR #294, opened 2025-09-30
fix: prevent crash during debug on fast view churn (like with HMR)

NathanWalker:
This case surfaced when doing fast HMR changes with Vite. During debug/development, these are benign and just due to race condition on js objects outliving their native wrapper very briefly. This just allows for quality of life debugging with fast HMR changes. Standard assert behavior is unchanged for release builds.

The small adjustment on `IndexedPropertySetterCallback` is a small related issue but important bugfix.

Original code: 
`if (wrapper == nullptr && wrapper->Type() != WrapperType::ObjCObject) { return; }` 

This is problematic because:

It can dereference wrapper when it’s null (the RHS of && is evaluated when the LHS is true).
The logic there was wrong: it only returns when both are true. If wrapper is non-null but not an ObjCObject, it wouldn’t return and would proceed with an incompatible wrapper.
Fixed code: `if (wrapper == nullptr || wrapper->Type() != WrapperType::ObjCObject) { return; }` 
This is the intended guard:

- Short-circuits safely when wrapper is null.
- Returns for any non-ObjCObject wrapper type.
- It also matches the getter’s guard just above, which already uses the same pattern. So this change prevents a possible null deref and aligns setter/getter behavior. It doesn't break existing behavior and makes the setter correctly bail out in the same cases the getter does.

### DIFF
--- a/NativeScript/runtime/ArgConverter.mm
+++ b/NativeScript/runtime/ArgConverter.mm
@@ -1,5 +1,6 @@
 #include <Foundation/Foundation.h>
 #include <sstream>
+#include <unordered_set>
 #include "ArgConverter.h"
 #include "NativeScriptException.h"
 #include "DictionaryAdapter.h"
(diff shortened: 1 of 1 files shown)


edusperoni:
this is definitely not a benign thing haha. Something that would usually crash/error out now silently fails like this and the developer has no idea what happened.

In my current case, my native object was being released prematurely and instead of getting a normal error I now get incomplete data:

call:
```
responseChunks.appendData(data);
```

instead of erroring on this, responseChunks has just been released and I get a very non-descriptive:

```
Note: ignore method on non-native receiver (class: NSMutableData, selector: appendData:, jsName: appendData, args: 1). Common during HMR.
```

PR #311, opened 2025-11-02
fix: native methods expecting a NSError arg will now throw a JS exception if the error arg is not passed

farfromrefug:
This is what we discussed on Discord.
If a native obj-c method like this `openReturningError(error?)` is called without the error argument it will throw a JS Error if an error is "triggered"
The JS Error named "NSError" will have a `localizedDescription` as the message and code,domain,nativeException properties.

@edusperoni dont hesitate to comment on the actual code. I am no v8 expert and i was helped by AI.
I tested the code and it works.

### DIFF
--- a/TestRunner/app/tests/ApiTests.js
+++ b/TestRunner/app/tests/ApiTests.js
@@ -337,7 +337,7 @@ describe(module.id, function () {
             expect(e.nativeException).toBeDefined();
             // The wrapped object should behave like an NSError proxy/wrapper
             // (we assert existence of localizedDescription property)
-            expect(typeof e.nativeException.localizedDescription === "string" || e.nativeException.localizedDescription instanceof String).toBe(true);
+            expect(typeof e.nativeException.localizedDescription).toBe('string');
         } finally {
             expect(isThrown).toBe(true);
         }


NativeScript/runtime/Interop.mm
```diff
@@ -1523,7 +1523,38 @@ inline bool isBool() {
...
+            // Attach the NSError code (number) and domain (string)
+            jsErrObj->Set(context, tns::ToV8String(isolate, "code"), Number::New(isolate, (double)[error code])).FromMaybe(false);
+            if (error.domain) {
+                jsErrObj->Set(context, tns::ToV8String(isolate, "domain"), tns::ToV8String(isolate, [error.domain UTF8String])).FromMaybe(false);
+            } else {
+                jsErrObj->Set(context, tns::ToV8String(isolate, "domain"), Null(isolate)).FromMaybe(false);
+            }
+
+            // Wrap the native NSError instance into a JS object and attach as nativeException
+            ObjCDataWrapper* wrapper = new ObjCDataWrapper(error);
```
edusperoni: This needs to be registered in the object manager otherwise it'll leak once the error is released

PR #315, opened 2025-11-20
feat: improve initialization with RAII patterns and runtime IDs

adrian-niculescu:
This PR improves runtime initialization correctness and modernizes V8 API usage:

**Context Management (RAII):**
Adds `Context::Scope` RAII pattern to `RunMainScript` for proper context handling. The `Context::Scope` ensures the context is correctly exited when the scope ends, even if exceptions occur. This is the recommended V8 C++ API pattern per `v8-locker.h` documentation.

Note: `Runtime::Init` continues to use manual `context->Enter()` because the context must remain entered for the runtime's entire lifetime, not just during initialization.

**Runtime ID Generation:**
Introduces a static atomic counter (`nextRuntimeId_`) that generates unique IDs for each `Runtime` instance during initialization. This provides reliable runtime identification for both main and worker threads using thread-safe atomic operations with `memory_order_relaxed` (appropriate for ID generation).

The patterns align with modern C++ best practices and V8 API recommendations.

### DIFF
--- a/NativeScript/runtime/Runtime.h
+++ b/NativeScript/runtime/Runtime.h
@@ -1,6 +1,8 @@
 #ifndef Runtime_h
 #define Runtime_h
 
+#include <atomic>
+
 #include "Caches.h"
 #include "Common.h"
 #include "MetadataBuilder.h"
@@ -59,6 +61,7 @@ class Runtime {
   static thread_local Runtime* currentRuntime_;
   static std::shared_ptr<v8::Platform> platform_;
   static std::vector<v8::Isolate*> isolates_;
+  static std::atomic<int> nextRuntimeId_;
   static SpinMutex isolatesMutex_;
   static bool v8Initialized_;
   static std::atomic<int> nextIsolateId;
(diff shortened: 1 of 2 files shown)


adrian-niculescu:
Noticed the test failure now. On it.

adrian-niculescu:
> The runtime ID change introduces a new bug where the main thread runtime is now considered a Worker. Also, you'll get duplicate runtime ids because the worker ids and runtime ids aren't generated from the same atomic int.
> 
> https://github.com/NativeScript/ios/blob/fd2703d59e6ba2472d8baefe014f0afc4c8952df/NativeScript/runtime/Worker.mm#L127
> 
> @NathanWalker please hold off merging until at least a proper review comes in.
> 
> I'm reverting this for now

@edusperoni, I apologize! You are right that the change is wrong in the current implementation due to `IsRuntimeWorker()` returning true for an id > 0.
I want to add that the change was not completely idiotic, just misplaced (should not have pushed it here): our fork that implemented support for ESM modules before you (you just recently did this in the 9.0.0 release) because we wanted to use dynamic imports; this was at a time when you didn't need this as bundling takes away the need for most situations where an import is needed. In our implementation the runtime id was used as a map key in the ESModule loader to ensure each runtime in the process gets its own one-time ESM setup. Sorry, I will come back with a cleaned-up PR with the other part of this PR which is legitimate.

PR #316, opened 2025-11-21
fix: prevent global Promise overwrite for thread safety.

adrian-niculescu:
This PR ensures thread safety for Promises by preventing the global `Promise` object from being overwritten by user-space code or 3rd party libraries.

NativeScript relies on a custom Promise proxy implementation to ensure that callbacks are executed on the same thread where they were created. This is critical for iOS, where UI updates must happen on the main thread.

Some JavaScript libraries (e.g., Mocha) attempt to replace the global `Promise` with their own implementation or polyfill. If this happens, the runtime's thread marshalling logic is lost, leading to deadlocks, race conditions, or crashes when callbacks run on the wrong thread.

This change wraps `globalThis` in a Proxy during initialization to intercept and silently block any attempts to redefine or assign to the `Promise` property, strictly preserving the runtime's thread-safe implementation.

Note: This has been battle-tested since Nov 2020 in a project I work on.

### DIFF
--- a/NativeScript/runtime/PromiseProxy.cpp
+++ b/NativeScript/runtime/PromiseProxy.cpp
@@ -72,6 +72,27 @@ void PromiseProxy::Init(v8::Local<v8::Context> context) {
                     });
                 }
             });
+            globalThis = new Proxy(globalThis, {
+                defineProperty: (target, prop, descriptor) => {
+                    if (prop === 'Promise') {
+                        return true;
+                    }
+                    return Object.defineProperty(target, prop, descriptor);
+                },
+                set: (target, prop, value, receiver) => {
+                    if (prop === 'Promise') {
+                        return true;
+                    }
+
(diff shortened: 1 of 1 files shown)


adrian-niculescu:
Closing this. Your main point stands: locking `globalThis.Promise` removes agency that libraries like zone.js rely on, so the runtime isn't the right place to enforce it.

On the overkill angle, one clarification, and it actually argues for dropping this: the proxy only traps `set` and `defineProperty` (no `get`), and bare globals and `global.x` resolve against the real global object rather than through it. So it doesn't slow down general global access, it only catches an explicit `globalThis.x =` write. The flip side is that `global.Promise =` or a plain `Promise =` bypass it completely, so it's a leaky guard as much as a heavy one.

What triggered this on our end was a test harness (mocha) replacing the global Promise, which is a test setup concern on our side. We'll handle it there instead.

Thanks for the review.

PR #319, opened 2025-11-25
fix: node built-in modules handling

NathanWalker:
- Node.js built-in modules (e.g., `node:url`) are now resolved by creating in-memory polyfill modules, registered and reused in a virtual registry, instead of writing polyfill files to disk. 
- Added new tests to verify that built-in modules like `node:url` and optional modules are resolved correctly using the new in-memory approach. 
- The `pathToFileURL` polyfill now encodes paths more accurately by replacing `%2F` with `/`, ensuring correct URL formatting.


simensBat:
This change helped me.

PR #353, opened 2026-04-03
revert: "fix: prevent crash during debug on fast view churn (like with HMR)"

edusperoni:
Reverts NativeScript/ios#294

This "fix" gets rid of an important part of the runtime stability where you can do something like:

```
const myNativeClass = SomeNSClass.new();
// myNativeClass is weakly held and we make SomeNSClass weak and it gets released
const value = myNativeClass.someNativeMethodThatAlwaysReturns(); // no error, just returns undefined
```

when this happens the error will not be thrown and cannot be caught, so if it happens you NEED to be looking at the console because it won't be caught by sentry/crashlytics/etc in production.

### DIFF
--- a/NativeScript/runtime/ArgConverter.mm
+++ b/NativeScript/runtime/ArgConverter.mm
@@ -7,7 +7,6 @@
 #include "Interop.h"
 #include "Helpers.h"
 #include "Runtime.h"
-#include "RuntimeConfig.h"
 
 using namespace v8;
 using namespace std;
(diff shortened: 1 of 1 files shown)


NathanWalker:
Consider this one which just improves it: https://github.com/NativeScript/ios/pull/354
that way we can still get best of both solutions.

NathanWalker:
@copilot compare this revert to this pull request: https://github.com/NativeScript/ios/pull/354

NathanWalker:
That was weird ed, I asked copilot to compare the revert to the other PR and it auto applied a change here? That was unexpected - I guess just see the comments and see what ya think. We can go full revert or just go with the improvement.

NathanWalker:
@copilot I'm not sure why you made changes to this PR, it was asked to just compare the revert to another pending PR. Ideally please take this PR back to it's original revert (the first commit in this PR) and just provide the analysis comparing the 2 prs. Update the title of this PR to what it was originally as well.

NathanWalker:
we'll go with improvement here: <https://github.com/NativeScript/ios/pull/354>

PR #354, opened 2026-04-03
fix: throw JS exception instead of silent warning for disposed native object calls in debug mode

NathanWalker:
- Replaces silent `Log()` warnings with proper `isolate->ThrowException()` calls when
  invoking methods on disposed native objects in debug mode
- Provides specific, actionable error messages that name the method, class, and selector
- Still avoids crashing the app in debug mode

## Context

The change in #294 (42a5328) prevented crashes during HMR (debug mode only) by replacing `tns::Assert` with
a `Log()` warning that silently returned `v8::Undefined`. While this fixed the crash, it
introduced a regression: errors became undetectable from JS, and `undefined` values could
silently leak into arrays and other data structures.

This fix takes the middle path — a catchable JS exception that:
1. Doesn't crash the app (HMR-safe)
2. Is detectable via `try/catch`
3. Prevents silent `undefined` propagation
4. Gives developers a clear message about what went wrong

### DIFF
--- a/NativeScript/runtime/ArgConverter.mm
+++ b/NativeScript/runtime/ArgConverter.mm
@@ -40,8 +40,7 @@ Local<Value> ArgConverter::Invoke(Local<Context> context, Class klass, Local<Obj
                 std::string errMsg = std::string("Cannot call method '") + jsNameStr +
                     "' on a disposed native object (class: " + classNameStr +
                     ", selector: " + selectorStr + "). This can happen during HMR or fast view churn.";
-                isolate->ThrowException(Exception::Error(tns::ToV8String(isolate, errMsg)));
-                return v8::Undefined(isolate);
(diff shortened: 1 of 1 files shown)


edusperoni:
Have you tried a throw NativeScriptException instead? Might still work and and it's a normal exception that can be caught

NathanWalker:
Good call @edusperoni, just pushed up to do standard `throw NativeScriptException`. It will work fine. 

The only downside for HMR workflows is that it'll be recommended to enable `discardUncaughtJsExceptions: true` during development (with hmr) since it's still likely superfluous native instance lifecycle errors can throw however they are harmless in these hmr workflows so enabling in nativescript.config is good solution for workflows encountering those or at minimum let's user decide what they prefer. This should be good to go now.

triniwiz (approved):

PR #375, opened 2026-05-17
fix: stop discarding errors in debug

edusperoni:
turns out I missed a few errors in my other PR.

Yes the app should crash in debug. Right now it crashes in release mode but continues fine in debug (?????????????)

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed a debug-only mechanism that could keep the app running after JS errors; unrecoverable errors now terminate as expected.
  * Worker error propagation tightened so uncaught worker errors are delivered reliably to the main runtime.

* **Refactor**
  * Module loading now surfaces exceptions immediately instead of attempting debug-only graceful continuation.
  * Module resolution, dynamic import handling, and diagnostic logging improved for clearer, more consistent failure messages.


NathanWalker:
@coderabbitai review
Also consider whether this adversely affects the `nativescript.config.ts` > `showErrorDisplay: true` option.

NathanWalker:
@copilot resolve the merge conflicts from Worker.mm

PR #382, opened 2026-06-03
fix(ios): NSError** runtime fallback and enriched arg-count diagnostics

NathanWalker:
- improved error visibility around NSError** out-parameter handling 

Runtime:
- Metadata.h: hasErrorOutParameter() now falls back to inspecting the last binary type encoding when the MethodHasErrorOutParameter flag is clear. Mirrors what ArgConverter::IsErrorOutParameter already does on the callback path. Lets apps shipping older/buggy metadata recover correct auto-NSError** handling without regenerating metadata.

Diagnostics:
- ArgConverter.mm: argument-count error message had its Actual/Expected labels swapped and no call-site context. Fixed the swap and added method kind, JS name, class name, full Objective-C selector, and a conditional hint when the method has an NSError** out-parameter that callers may omit.
- ObjectManager.mm: ReleaseNativeCounterpartCallback error message now names the function it is about (__releaseNativeCounterpart).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messaging for argument-count mismatches, now correctly reporting actual vs. expected parameter counts with additional context about method type and class information.
  * Enhanced error parameter detection for more reliable method signature validation across modern SDK annotations.

### DIFF
--- a/NativeScript/runtime/ArgConverter.mm
+++ b/NativeScript/runtime/ArgConverter.mm
@@ -104,9 +104,24 @@ Local<Value> ArgConverter::Invoke(Local<Context> context, Class klass, Local<Obj
   if ((!meta->hasErrorOutParameter() && args.Length() != argsCount) ||
       (meta->hasErrorOutParameter() && args.Length() != argsCount &&
        args.Length() != argsCount - 1)) {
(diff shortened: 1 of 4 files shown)


NathanWalker:
No longer needed.

PR #383, opened 2026-06-07
feat: ESM resolver hardening, HTTP module loader, ns:module dev surface

NathanWalker:
Framework-agnostic hot module replacement on iOS with native ES modules: the device fetches modules over HTTP from the Vite dev server and applies hot updates without restarting the process.

The runtime's entire dev surface is one builtin module — `ns:module` — resolved through the same `ns:` registry as `ns:util` (#418): `require("ns:module")`, static `import`, and `import()` all yield the same frozen per-realm module, materialized lazily on first resolution. The dev surface defines no globals. Five primitives, each traceable to a V8-embedder or OS constraint:

| Member | Why it must be native |
|---|---|
| `configureLoader(config)` | Import map, volatile URL patterns, and the canonicalization vocabulary, consumed inside V8's synchronous `ResolveModuleCallback`. The sole channel by which server/framework URL policy enters the runtime — native code carries no URL vocabulary of its own |
| `invalidateModules(urls)` | Drops host-owned `v8::Module` records and arms a CFNetwork cache-bust nonce |
| `getLoadedModuleUrls()` | Registry introspection for the JS full-reload path |
| `createRequire(filenameOrURL)` | Mints a require bound to a base directory with the strict evaluation policy — policy selection and the require factory are embedder machinery |
| `createPumpingRequire(filenameOrURL)` | Same, but async graphs are driven to settlement on nestable v8 platform tasks — only the embedder can pump its own event loop |

A new `node:module` builtin re-exports `createRequire` only, so `import { createRequire } from 'node:module'` works unmodified (no `.d.ts` is shipped for it — `@types/node` already declares that ambient module). Debug builds also carry `canonicalizeHttpUrlKey(url)`, a pure test diagnostic. Missing members are simply absent — never present-but-throwing — so feature checks work.

## Module-graph pipeline

Every ES module root — boot entry, `require()`, dynamic `import()` — runs the same discovery pipeline (`StartModuleGraphLoad`) before instantiation, dispatching per edge: local files read and compile inline during discovery, HTTP edges fetch concurrently on `NSURLSession` background queues, and both register under the canonical keys the resolver uses — resolution parity is guaranteed by one shared `ResolveSpecifierToPath` rather than by convention. Instantiation then runs with a lookup-only synchronous `ResolveModuleCallback`; evaluation is promise-chained under top-level await. Mixed disk↔HTTP graphs fetch their HTTP members in parallel; a disk-only graph completes discovery inline and pays no pump. A synchronous fetch survives only as a loudly-logged anomaly guard for edges the walk could not resolve (it logged zero hits across the suite). Boot pressure is answered at the source: the dev server pre-bundles `@nativescript/core` and node_modules into single-eval payloads, and the pipeline fetches the remaining app graph concurrently.

Responses are judged once, by a shared classifier, for both transports: a JavaScript MIME essence loads as JS, a JSON MIME loads as a JSON module, and 204/205, missing, or foreign MIME types are fetch failures naming the URL and the received type — the web's module-script rule, so a dev server's SPA fallback answering `200 text/html` rejects with the real reason instead of compiling HTML as JavaScript. An empty 2xx JavaScript body stays a valid empty module (type-only TS modules transform to zero code).

## Module identity & freshness

Module identity is the canonical URL: the server emits exactly one URL per module and never varies it for freshness (this closes the realm-split / `Cannot redefine property` crash class). Canonicalization survives only to absorb externally-caused variance (Vite's `?v=`/`?import`/`?t=` markers, `file://http://` wrapping); the *mechanism* (fragment strip, param drop, sort) is native, while the *vocabulary* (which params to strip, which path prefixes are dev endpoints, which paths keep their query verbatim) is supplied by the client via `configureLoader`. Freshness is explicit eviction at both layers that could serve a stale byte: the V8 module registry, and a one-shot `__ns_dev_nonce` that defeats CFNetwork's cache (observed serving stale bodies on iOS 18+/26 Sim despite `no-store` and a zero-capacity `NSURLCache`).

## Loader architecture

Disk and HTTP ES modules now share one architecture, aligned with the invariants Node and Blink both obey: the resolver only ever compiles and registers — it never instantiates or evaluates — so V8 drives graph discovery during the root's `InstantiateModule`, evaluation happens once at the root in spec order, and import cycles terminate through the registry the way module-map self-inserts do elsewhere. The re-entry/fallback/gating machinery that previously compensated for resolver-order evaluation is deleted. Per-isolate loader state (registry, in-flight bookkeeping, waiters, async graph loads) lives in a typed `Caches` state slot destroyed with the isolate, replacing process-wide isolate-keyed maps; module→key reverse lookups (referrer discovery, `import.meta`) go through an identity-hash index instead of registry scans. Boot state is derived natively — the cold-boot runloop pump is armed only while an entry module is evaluating on that thread — which is why `setDevBootComplete` no longer exists, and the module-graph waits share one deadline constant (transport < settle < boot backstop, by construction).

## Module evaluation modes & require(esm)

Graph evaluation is one primitive, `EvaluateModuleGraph`, under three named policies:

- **`kSyncStrict`** — Node's `require(esm)`: `IsGraphAsync()` is refused *before* evaluation ever starts (the module stays instantiated and loadable via `import()`, and the refusal wins over any runtime error the graph would have produced), including on registry hits already at `kEvaluated` — a TLA-parked graph reports `kEvaluated` with a pending capability promise, so returning its namespace would hand `require()` a TDZ binding. Sync graphs must settle synchronously; the promise state is read directly.
- **`kSyncPumping`** — the boot/worker-entry settle loop: nestable v8 tasks + microtask checkpoints, HTTP-only runloop slices, 1s local yield window vs 60s HTTP deadline.
- **`kAsync`** — evaluate and hand back the capability promise (the dynamic-import shape).

`require()` of an ES module is now **Node-strict** (previously it fell into the pumping path by accident): a graph containing top-level await throws, matching `ERR_REQUIRE_ASYNC_MODULE`. Its exports follow Node's `populateCJSExportsFromESM` exactly — a literal `'module.exports'` export wins; a namespace with no default or its own `__esModule` passes through; otherwise a synthetic facade module re-exports the target with live bindings and `__esModule = true`, so transpiled consumers (`_mod.__esModule ? _mod.default : _mod`) find the real default. `createPumpingRequire` is the explicit escape hatch for native-boundary calls: async graphs are driven to settlement (60s deadline, throws rather than returning a half-initialized namespace) without ever advancing the Cocoa runloop — outside boot the runloop belongs to the app. Also fixes `require("~foo")`, which previously lost its first letter to an unconditional `substr(2)`.

## Additional

- ESM resolver hardening (`ModuleInternalCallbacks.mm`): HTTP(S) URLs end-to-end (resolve, fetch, dynamic import) and `.json` imports compiled into synthetic ES modules. Builtin specifiers (`ns:`, registered `node:`) resolve only through the registry — an import-map entry can never shadow them onto HTTP.
- Worker correctness: module registries keyed by `v8::Isolate*` (not thread-locals); worker teardown preserves the main isolate's process-wide dev state; worker entry-script errors propagate to `worker.onerror`.
- Security: every remote fetch — the sync fallback and the async graph walk alike — passes `IsRemoteUrlAllowed()` (`DevFlags.mm`): deny-by-default in release, opt-in via `security.allowRemoteModules` (+ optional `remoteModuleAllowlist`).
- On-device tests (`TestRunner/app/tests/HttpEsmLoaderTests.js`, `EsmInteropTests.js`, `CreateRequireTests.js`) pin the `ns:module` module shape, the `require`/`import()` identity, the exports-interop cascade, and the strict/pumping differential.


NativeScript/runtime/URLImpl.cpp
```diff
@@ -18,6 +18,88 @@ void URLImpl::Init(v8::Isolate* isolate, v8::Local<v8::ObjectTemplate> globalTem
     globalTemplate->Set(urlPropertyName, URLTemplate);
 }
 
+void URLImpl::InstallBlobMethods(v8::Local<v8::Context> context) {
```
NathanWalker: Blob can likely be split into separate PR. Not directly attributable to this pull request (at a time, I used blob urls with hmr updates but found standard http networking of es modules to suffice fine) - but useful on it's own.

edusperoni:
I'm not sure I understand the goal of these "dev sessions"? What do these need specifically from the runtime that's not an "user land" thing?

NathanWalker:
> I'm not sure I understand the goal of these "dev sessions"? What do these need specifically from the runtime that's not an "user land" thing?

Yeah good question @edusperoni, the "dev session" naming here probably over (or mis) characterizes things.

The "session" part is just the contract for booting from a dev server over HTTP instead of the on-disk bundle. It does three things a one-shot bundle loader doesnt: point resolution at an HTTP origin + install the import map before the first import, give a re-entrant boot (import client > import entry) that can re-run for a full reload without relaunching the process, and bubble import failures back as a rejected promise so the client can show an overlay instead of the app just dying.

Could it be userland? I think the thing that trips people up (tripped me up too) is that on the web HMR is userland because the browser is the runtime; it already ships a spec ESM loader that fetches over HTTP and a host-owned module map you poke at by varying the URL. Vite's client gets to be "just JS" because it sits on top of that. Here V8 is embedded by us, and bare V8 ships no loader at all; every piece of it is an embedder host callback only native can install. So the litmus test is pretty clean: anything that has to install/drive a V8 host callback or mutate V8's module map cant be userland, everything else stays in JS.

This may help expand a few things:
- The HTTP ESM loader, `import 'http://...'`, is resolved inside ResolveModuleCallback / the dynamic-import callback, before any user code for that module even runs. And in V8 10.3.22 that static callback is synchronous, which is the whole reason the prefetch engine exists (issue with V8's sync walk from JS, and a serial network walk on the UI thread trips the launch watchdog; an iOS issue I sort of knew about long ago so was interesting to be reminded of here).
- Identity-preserving eviction (`__nsInvalidateModules`), the registry is host-owned (`g_moduleRegistry`) and theres no JS API to evict/re-instantiate a compiled record. This is the most NS-specific one: the web just mints a new identity per save with `?t=`, but we cant, because module identity is load-bearing for native interop - mint a fresh realm for a @nativescript/core module and the native patches collide leading to `Cannot redefine property`. So the runtime has to own eviction that collapses back to one canonical identity.
- import.meta.hot: import.meta only gets populated in `SetHostInitializeImportMetaObjectCallback`, and hot.data only survives a swap if it's keyed to the runtime's canonical module key, which userland doesnt have.
- Reboot teardown: terminating native-owned workers + ordered v8::Global/thread-local cleanup before isolate disposal.
Everything protocol-ish I deliberately left in JS: the WebSocket, the wire format, evictPaths/closure computation, accept/dispose policy; all in @nativescript/vite's client. Runtime exposes mechanism, client owns policy; nothing links or version-pins vite, and unknown URLs just fall through to a generic HTTP loader.

So really these globals arent "a dev-session feature" so much as the embedder half of a spec ESM loader + an identity-preserving module map. The part the browser hands Vite for free. There is likely a few that could move to JS if we want a smaller surface like __nsApplyStyleUpdate (just Application.addCss + restyle) and __nsGetLoadedModuleUrls (introspection) and others. Lmk if you see the boundary differently and we can make further adjustments.

PR #384, opened 2026-06-07
feat: reloadApplication for JS bundle restart without restarting app process

NathanWalker:
```
NativeScriptRuntime.reloadApplication()
```

Helpful for programmatic reset of JS isolate for clean restart of JS application as well as OTA (over-the-air) updates without restarting the entire app process.

https://github.com/NativeScript/NativeScript/pull/11261

## Summary by CodeRabbit

* **New Features**
  * Added application reload support through the native runtime API.
  * Reloads preserve the main runtime, isolate, native delegates, and shared application instance.
  * Supports optional base directories for loading updated modules.
  * Exposes a reload count and triggers an application reload callback.

* **Bug Fixes**
  * Improved callback safety when the runtime or isolate is unavailable.
  * Clears loaded modules and pending runtime state during reloads.

* **Tests**
  * Added coverage for reload behavior, module re-evaluation, lifecycle callbacks, and state preservation.

### DIFF
--- a/NativeScript/NativeScript.h
+++ b/NativeScript/NativeScript.h
@@ -26,3 +26,10 @@
 - (bool)liveSync;
 
 @end
+
+@interface NativeScriptRuntime : NSObject
+
++ (BOOL)reloadApplication;
++ (BOOL)reloadApplication:(NSString*)baseDir;
+
+@end
--- a/NativeScript/NativeScript.h
+++ b/NativeScript/NativeScript.h
@@ -16,6 +16,11 @@
 
 - (instancetype)initWithConfig:(Config*)config;
 - (void)runScriptString: (NSString*) script runLoop: (BOOL) runLoop;
+
+/**
+ Embedder-only: dispose the isolate and construct a new Runtime.
+ JS-bootstrapped apps must not use this — it kills JS-backed delegates.
+ */
 - (void)restartWithConfig:(Config*)config;
 - (void)shutdownRuntime;
 
@@ -29,6 +34,13 @@
 
 @interface NativeScriptRuntime : NSObject
 
+/**
+ Reset the JS application on the existing main isolate.
+ Does not dispose the isolate, create a second Runtime, or re-enter
+ UIApplicationMain. JS-backed AppDelegate / UIScene delegates stay alive.
+ Optional baseDir points later module loads at an OTA bundle. After flushing
+ module caches, invokes global.__onApplicationReload if present.
+ */
 + (BOOL)reloadApplication;
 + (BOOL)reloadApplication:(NSString*)baseDir;
 
--- a/NativeScript/NativeScript.mm
+++ b/NativeScript/NativeScript.mm
@@ -21,6 +21,21 @@ namespace tns {}
 
 @end
 
+static Config* CopyConfig(Config* config) {
+  Config* copy = [[Config alloc] init];
+  copy.BaseDir = config.BaseDir;
+  copy.ApplicationPath = config.ApplicationPath;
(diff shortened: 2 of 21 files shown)


edusperoni:
I don't see this ever being a good idea.

Bundle 1 calls UIApplicationMain with a custom AppDelegate extended in JS
Bundle 1 calls reloadApplication, destroying its isolate
Bundle 2 loads. AppDelegate is inactive (was killed with previous isolate), any methods called will silently no-op

NathanWalker:
> I don't see this ever being a good idea.
> 
> Bundle 1 calls UIApplicationMain with a custom AppDelegate extended in JS Bundle 1 calls reloadApplication, destroying its isolate Bundle 2 loads. AppDelegate is inactive (was killed with previous isolate), any methods called will silently no-op

Yeah, the delegate class outlives the isolate, UIKit keeps dispatching to the old instance, and MethodCallback bails when the isolate's gone. So a JS-extended AppDelegate from bundle 1 would go inert after reload.

The existing teardown behavior (same IsolateWrapper checks in ClassBuilder, the adapters, etc), and restartWithConfig itself has been in the runtime since the embedding work in [#231](https://github.com/NativeScript/ios/pull/231). This just makes it callable from JS with the original config kept around.

The reason it holds up in practice is that core doesn't actually rely on delegate dispatch for lifecycle, everything goes through NSNotificationCenter observers that get re-registered when the new isolate boots, and the soft-reboot path recreates the window against the live scene. We've been shipping these PR changes in production this year and it's been solid on this path, so this is proven useful already.

The case you're describing (custom delegate methods like push token callbacks or openURL) is something worth handling though. Two things I can do as follow-up: have the soft-reboot path reassign UIApplication.sharedApplication.delegate to a fresh instance from the new bundle (with a strong ref since UIKit won't hold one), and add a log behind the script-loading flag when a dead-isolate callback fires so this failure mode is visible instead of silent.

If you can think of other dispatch paths that'd stay pinned to the old isolate, lemme know.

edusperoni:
For now. With UIScene we now use delegates, so this only works until iOS 27. It works in embedding because in embedded we don't bootstrap the app and just render views on existing apps

NathanWalker:
Thanks again for all the feedback @edusperoni, I've pushed up the follow-ups to address all cases (modern UIScene delegates and soon to be legacy standard delegate behavior).
see https://github.com/NativeScript/ios/pull/384/changes/fc0e142454498fbd91d147a65022ff988c978edf
And also the core adjustment to support both from the supporting pr: https://github.com/NativeScript/NativeScript/pull/11261/changes/6c1479b255b82cd77e674a27ec2ce35a9a4bc3e7

edusperoni:
I still don't understand why this is needed. Is this for HMR? Because it seems like we're replacing irreplaceable parts of the application. Like if you change the app bootstrap or things that touch delegates we shouldn't expect the HMR to kick in instead of a full restart.

Extended classes are dynamic already, so even if we reevaluate modules they'd just create new native classes that would get picked up when a new view is created.

What does this bring in? Like 2s when someone touches critical files?

NathanWalker:
This is for seamless production OTA (over-the-air) updates (unrelated to hmr).

edusperoni:
For OTA this is even worse! Half of people's delegates will be gone by that point and callbacks likely will just leak all around from the previous isolate. I could understand this for dev, but for prod it's just asking for crashes and weird behavior that are pretty much unfixable. Things like sentry and firebase will likely not work properly due to trying to initialize it multiple times for example

NathanWalker:
This has already been in production for 6 months and proven stable, it's very helpful. I can show you offline how it works - will share details later this week.

PR #392, opened 2026-06-13
Make V8 string to NSString conversions UTF-16 faithful

adrian-niculescu:
V8 strings are UTF-16, but a few of the V8 <-> NSString bridge points round-trip through UTF-8 first, and that loses data in two ways:

- **Lone surrogates** get corrupted. A JS string can hold an unpaired surrogate (for example `"\uD834"`), which has no valid UTF-8 encoding, so the round-trip replaces it with U+FFFD. This affects every site that goes through the UTF-8 path, including `tns::ToUtf16String` itself.
- **Embedded NUL** gets truncated at the sites that use C-string APIs. `DictionaryAdapter` builds strings with `[NSString stringWithUTF8String:result.c_str()]`, and a couple of reverse-direction spots fold `[obj UTF8String]` into a `std::string`, both of which stop at the first `\0`.

The fix is to keep everything in UTF-16:

- `tns::ToUtf16String` now reads the V8 string's native two-byte buffer directly via `v8::String::Value` instead of `ToString` + `std::codecvt_utf8_utf16`. As a bonus that drops the deprecated `codecvt` (there was already a `// FIXME` on it) and is faster.
- The JS -> NSString sites in `DictionaryAdapter`, `Interop` and `ArgConverter` now build the NSString with `stringWithCharacters:length:` off `ToUtf16String`, instead of `ToNSString` / `stringWithUTF8String:`.
- The two NSString -> V8 spots (`DictionaryAdapter objectForKey:` and the indexed getter in `ArgConverter`) pass the `NSString` straight to `ToV8String` instead of going through a `UTF8String` C string. `ToV8String(NSString*)` already exists and uses an explicit byte length, so it is NUL safe.

I left `tns::ToVector` alone on purpose. Its input is already a UTF-8 `std::string`, so it is a different conversion and not part of this bug.

Test: added a `TestRunner` case that bridges a JS string with a lone high surrogate to an NSString and checks the code unit survives. Without the fix it comes back as U+FFFD.

Formatting note: I matched each file's existing indentation (`DictionaryAdapter.mm` is 4-space, the others are 2-space) and did not run clang-format, to keep the diff to just the real changes.

We have shipped this in our downstream fork for a while.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved preservation of special Unicode characters (including surrogates and embedded NUL characters) during string conversions between JavaScript and native code.

* **Tests**
  * Added test coverage for special Unicode character handling to ensure correct preservation during conversions.

### DIFF
--- a/NativeScript/runtime/ArgConverter.mm
+++ b/NativeScript/runtime/ArgConverter.mm
@@ -294,7 +294,8 @@ void ArgConverter::SetValue(Local<Context> context, void* retValue, Local<Value>
   } else if (value->IsString()) {
     if (type == BinaryTypeEncodingType::IdEncoding ||
         type == BinaryTypeEncodingType::InterfaceDeclarationReference) {
-      id data = tns::ToNSString(isolate, value);
+      std::u16string strValue = tns::ToUtf16String(isolate, value);
+      id data = [NSString stringWithCharacters:(const unichar*)strValue.data() length:strValue.size()];
       // this feels wrong but follows the other CFBridgingRetain calls
       // and also solves a leak
       auto ref = CFBridgingRetain(data);
@@ -929,8 +930,8 @@ void ArgConverter::IndexedPropertyGetterCallback(uint32_t index,
   }
 
   if ([obj isKindOfClass:[NSString class]]) {
-    const char* str = [obj UTF8String];
-    args.GetReturnValue().Set(tns::ToV8String(isolate, str));
+    NSString* nativeStr = (NSString*)obj;
+    args.GetReturnValue().Set(tns::ToV8String(isolate, nativeStr));
     return;
   }
 
--- a/NativeScript/runtime/ArgConverter.mm
+++ b/NativeScript/runtime/ArgConverter.mm
@@ -294,8 +294,7 @@ void ArgConverter::SetValue(Local<Context> context, void* retValue, Local<Value>
   } else if (value->IsString()) {
     if (type == BinaryTypeEncodingType::IdEncoding ||
         type == BinaryTypeEncodingType::InterfaceDeclarationReference) {
-      std::u16string strValue = tns::ToUtf16String(isolate, value);
-      id data = [NSString stringWithCharacters:(const unichar*)strValue.data() length:strValue.size()];
+      id data = tns::ToNSString(isolate, value);
       // this feels wrong but follows the other CFBridgingRetain calls
       // and also solves a leak
       auto ref = CFBridgingRetain(data);
--- a/NativeScript/runtime/DictionaryAdapter.mm
+++ b/NativeScript/runtime/DictionaryAdapter.mm
@@ -48,7 +48,8 @@ using namespace tns;
         bool success = array->Get(context, self->index_).ToLocal(&key);
         tns::Assert(success, isolate);
         self->index_ += 2;
-        NSString* result = tns::ToNSString(isolate, key);
+        std::u16string keyStr = tns::ToUtf16String(isolate, key);
+        NSString* result = [NSString stringWithCharacters:(const unichar*)keyStr.data() length:keyStr.length()];
         return result;
     }
 
@@ -116,8 +117,8 @@ using namespace tns;
         bool success = properties->Get(context, (uint)self->index_).ToLocal(&value);
         tns::Assert(success, isolate);
         self->index_++;
-        std::string result = tns::ToString(isolate, value);
-        return [NSString stringWithUTF8String:result.c_str()];
+        std::u16string result = tns::ToUtf16String(isolate, value);
+        return [NSString stringWithCharacters:(const unichar*)result.data() length:result.size()];
     }
 
     return nil;
@@ -139,8 +140,8 @@ using namespace tns;
         Local<Value> value;
         bool success = properties->Get(context, i).ToLocal(&value);
         tns::Assert(success, isolate);
-        std::string result = tns::ToString(isolate, value);
-        [array addObject:[NSString stringWithUTF8String:result.c_str()]];
+        std::u16string result = tns::ToUtf16String(isolate, value);
+        [array addObject:[NSString stringWithCharacters:(const unichar*)result.data() length:result.size()]];
     }
 
     return array;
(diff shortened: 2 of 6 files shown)


adrian-niculescu:
@coderabbitai thanks, but I'm going to skip this suggestion, it would actually break the test.

Completing the round trip with `ns.toString()` would fail: the reverse NSString to JS direction still goes through UTF-8 (`ToV8String(NSString*)` uses `[value UTF8String]`), so it turns the lone surrogate back into U+FFFD. `expect(ns.toString()).toBe("\uD834")` and `charCodeAt(0) === 0xD834` would both fail. I confirmed that by running the TestRunner suite locally.

That reverse path is intentionally not touched for surrogates in this PR. The reverse-direction edits here only fix NUL truncation (passing an explicit byte length instead of folding through a C string), not surrogate fidelity, so there is no faithful NSString to JS path to round-trip a lone surrogate through. That is exactly why the test observes the NSString with a numeric native query (`lengthOfBytesUsingEncoding:` is 0 for a faithfully kept lone surrogate, 3 once it has become U+FFFD) instead of converting back to a JS string.

The bidirectional NUL case is already covered by the existing `Marshals NSString with null character` test in `NSStringTests.js`.

adrian-niculescu:
> 🧹 Nitpick comments (1)
> > TestRunner/app/tests/ApiTests.js (1)> `15-31`: _⚡ Quick win_
> > > **Consider adding a lone low surrogate test case for completeness.**
> > > The current test verifies a lone high surrogate (U+D834, range 0xD800-0xDBFF). Adding a similar test for a lone low surrogate (e.g., `"\uDC00"`, range 0xDC00-0xDFFF) would strengthen coverage of the UTF-16 fidelity fix, since both surrogate ranges should be preserved but have different bit patterns.
> > > 
> > > 📝 Optional additional test case
> > > ```js
> > > it("preserves a lone low surrogate when bridging a JS string to NSString", function () {
> > >     var ns = NSString.stringWithString("\uDC00");
> > >     expect(ns.length).toBe(1);
> > >     
> > >     var buffer = interop.alloc(interop.sizeof(interop.types.uint16));
> > >     ns.getCharactersRange(buffer, NSMakeRange(0, 1));
> > >     var codeUnit = new interop.Reference(interop.types.uint16, buffer).value;
> > >     interop.free(buffer);
> > >     
> > >     expect(codeUnit).toBe(0xDC00);
> > > });
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@TestRunner/app/tests/ApiTests.js` around lines 15 - 31, The test suite
> > > currently only covers high surrogates (U+D834) but lacks coverage for low
> > > surrogates in the UTF-16 bridging behavior. Add a new test case after the
> > > existing "preserves a lone surrogate when bridging a JS string to NSString" test
> > > that follows the same pattern but uses a low surrogate code unit (U+DC00 instead
> > > of U+D834) to verify that lone low surrogates are also preserved correctly when
> > > bridging JS strings to NSString. The new test should verify that the bridged
> > > NSString preserves the low surrogate value as 0xDC00.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Nitpick comments:
> In `@TestRunner/app/tests/ApiTests.js`:
> - Around line 15-31: The test suite currently only covers high surrogates
> (U+D834) but lacks coverage for low surrogates in the UTF-16 bridging behavior.
> Add a new test case after the existing "preserves a lone surrogate when bridging
> a JS string to NSString" test that follows the same pattern but uses a low
> surrogate code unit (U+DC00 instead of U+D834) to verify that lone low
> surrogates are also preserved correctly when bridging JS strings to NSString.
> The new test should verify that the bridged NSString preserves the low surrogate
> value as 0xDC00.
> ```
> 
> ℹ️ Review info

@coderabbitai, re the low surrogate nitpick: added a lone low surrogate case (U+DC00) next to the high surrogate one in 8cc59bf, and renamed the original so the pair reads clearly. Both read the bridged code unit back as a number via getCharacters:range: instead of a UTF-8 conversion, since converting a lone surrogate to UTF-8 hangs the test runner on CI's iOS 17.5 simulator.

edusperoni:
I'm curious, should we not just use utf16 everywhere we can then? Also seems like we're doing extra steps there: first convert to std::utf16string by passing a buffer + length and then pass the copied buffer + length to NSString. I feel we could have a `tns::ToNSString` which just reads the utf16 buffer directly.

I believe most of the utf8 usages we have in this repo come from the misconception that v8 wasn't using utf16 internally or that the utf8 string conversion would be faster somehow (I know I myself already fixed a bunch of utf8 issues with null character in the middle of the string)

adrian-niculescu:
> I'm curious, should we not just use utf16 everywhere we can then? Also seems like we're doing extra steps there: first convert to std::utf16string by passing a buffer + length and then pass the copied buffer + length to NSString. I feel we could have a `tns::ToNSString` which just reads the utf16 buffer directly.
> 
> I believe most of the utf8 usages we have in this repo come from the misconception that v8 wasn't using utf16 internally or that the utf8 string conversion would be faster somehow (I know I myself already fixed a bunch of utf8 issues with null character in the middle of the string)

Indeed, V8 strings are UTF-16 the whole way down, so these UTF-8 round-trips are just leftover from the old assumption. Same root cause as the embedded-NUL cases you mentioned.

And you're right that we don't need a new helper. `tns::ToNSString(isolate, value)` already exists, I'd just stopped calling it because its body still went through `String::Utf8Value`. I made it read the two-byte buffer and hand it straight to `stringWithCharacters:length:`, then pointed the call sites back at it. That drops the intermediate `std::u16string` copy and keeps the conversion in one place. Pushed.

On utf16 everywhere: agreed. The spot this doesn't close yet is the reverse direction. `ToV8String(NSString*)` still goes through `[value UTF8String]`, so it's NUL-safe but not surrogate-safe, and there's already a commented-out `NewFromTwoByte` path sitting in there for it. Happy to finish that so the bridge is symmetric, reading with `getCharacters:range:` rather than the commented `getBytes:...NSUTF16StringEncoding` (which can write a BOM). A wider UTF-8 sweep across the repo is probably its own PR.

edusperoni (approved):

adrian-niculescu:
Thanks for taking a look at this!

PR #393, opened 2026-06-17
fix(http-loader): survive the Network permission prompt on first time physical install

NathanWalker:
- Strategy to aid the first time install of hmr dev sessions on physical devices which will do a 1 time network permission prompt before beginning the boot loader


NathanWalker:
Noting that this is not a guarantee and just a potential direction on handling. Something similar to this direction.may work. It some cases this would resolve it, others not so it's timing related largely.

NathanWalker:
Going to address this differently with latest revisions.

PR #395, opened 2026-06-24
feat: swift package release workflow

NathanWalker:
Build and release pipeline for `NativeScript.xcframework` runtime binary to no longer be bundled inside the npm packages. Instead, they are published as GitHub Release assets and referenced via Swift Package Manager (SPM) in the Xcode project templates. The npm packages now only contain the Xcode project template and the metadata generator. The workflow is restructured to support this new distribution model, with improved artifact handling and SPM manifest stamping.

**Required to merge first before this**: https://github.com/NativeScript/ios-spm/pull/2

## Summary by CodeRabbit

* **New Features**
  * Package templates now resolve SwiftPM binary artifacts (remote Swift package products) instead of embedding prebuilt xcframeworks.
  * Expanded release pipeline to build, checksum, publish, and stamp artifacts for both iOS and visionOS.
  * Added automation to stamp SwiftPM manifest/package and template runtime versions during packaging.

* **Bug Fixes**
  * Improved artifact consistency with per-target versioning, checksum generation, and a SwiftPM resolution smoke test.

* **Chores**
  * Refreshed build/release workflow structure to support updated version/tag handling and per-target publishing.

### DIFF
--- a/.github/workflows/npm_release.yml
+++ b/.github/workflows/npm_release.yml
@@ -9,13 +9,57 @@ env:
   NPM_TAG: "next"
   XCODE_VERSION: "^15.0"
 
+# The runtime xcframeworks are no longer shipped inside the npm packages. Each
+# build publishes them as GitHub Release assets and points the SwiftPM manifest
+# (github.com/NativeScript/ios-spm) at them via binaryTarget(url:checksum:). The
+# slim npm package only carries the Xcode project template + metadata generator.
+#
+# Flow: setup → build (ios, visionos) → test → github-release (assets) →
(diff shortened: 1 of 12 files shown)


edusperoni:
This just broke next builds and much more. Why has such a massive change been merged with no oversight?

NathanWalker:
This was merged for further verification on `next` tag (next is handled different than other prerelease manual workflow triggers so a full run through on next to verify fully). Alpha confirmed so functionally all operational (just a next tag complication to avoid overtaxing the build system when only one target is needed). this should handle `next` better: https://github.com/NativeScript/ios/pull/405

PR #396, opened 2026-06-25
fix: only marshal promise resolution when created on the runtime loop

### LINKED ISSUE #330: Promise never resolves when resolve() is called inside setTimeout from a background thread
A common JS pattern for small delays hangs indefinitely on iOS when the Promise is created on a background thread:

```javascript
async function doSomething() {
    // wait 25ms for some events to propagate
    await new Promise(resolve => setTimeout(() => resolve(), 25));

    console.log('never reached');
}
```

This happens when native code (e.g. a NativeScript plugin) invokes JS callbacks on a background thread.

## Reproduction

Any native plugin that calls into JS from a background thread can trigger this. Example with a WebSocket delegate:

```
Native WebSocket library (background thread)
  │
  ▼
Plugin's native callback handler
  │  (running on background thread)
  ▼
NativeScript bridge invokes JS callback
  │  (acquires V8 locker on this background thread)
  ▼
JS event handler runs
  │
  ▼
await new Promise(resolve => {
    setTimeout(() => {
        resolve();  ← never executes
    }, 25);
});
```

The Promise hangs forever.

## Cause

PromiseProxy.cpp captures `CFRunLoopGetCurrent()` when the Promise is constructed:
https://github.com/NativeScript/ios/blob/73332e13667730b3084aa3dd7973685196efa086/NativeScript/runtime/PromiseProxy.cpp#L16

If `resolve()` is later called from a different thread, it uses `CFRunLoopPerformBlock` to schedule the actual resolution on the original thread's runloop:
https://github.com/NativeScript/ios/blob/73332e13667730b3084aa3dd7973685196efa086/NativeScript/runtime/PromiseProxy.cpp#L36-L37

The problem: background threads from dispatch queues or thread pools have dormant runloops. No code is calling `CFRunLoopRun()` on them - they just execute tasks and return to the pool. The scheduled block never gets processed.

Meanwhile, `setTimeout` always fires on `Runtime::RuntimeLoop()`:
https://github.com/NativeScript/ios/blob/73332e13667730b3084aa3dd7973685196efa086/NativeScript/runtime/Timers.cpp#L125

Timers are added to that runloop here:
https://github.com/NativeScript/ios/blob/73332e13667730b3084aa3dd7973685196efa086/NativeScript/runtime/Timers.cpp#L252

This is typically the main thread (captured at Runtime creation), which is different from the background thread where the Promise was created.

The flow:

1. JS callback runs on background thread B (V8 locker acquired)
2. Promise created, PromiseProxy captures thread B's runloop
3. setTimeout schedules timer on Runtime's runloop (thread A)
4. Callback completes, thread B returns to pool (runloop dormant)
5. Timer fires on thread A
6. resolve() called on thread A
7. PromiseProxy sees different runloops, uses CFRunLoopPerformBlock to schedule on thread B
8. Thread B's runloop is dormant - nobody is pumping it
9. Block sits there forever, Promise never resolves

## Suggested fix

Instead of marshaling to the creation thread's runloop (which may be dormant), marshal to the Runtime's runloop which is always active:

```javascript
// In PromiseProxy.cpp JS source
let runtimeRunloop = __getRuntimeRunloop();  // new global exposing Runtime::RuntimeLoop()

origFunc(value => {
    if (isFulfilled()) return;
    let res = resolve;
    markFulfilled();

    if (runtimeRunloop === CFRunLoopGetCurrent()) {
        res(value);
    } else {
        CFRunLoopPerformBlock(runtimeRunloop, kCFRunLoopDefaultMode, () => res(value));
        CFRunLoopWakeUp(runtimeRunloop);
    }
}, ...);
```

This ensures Promise resolution always happens on the Runtime's thread (where timers fire), regardless of which thread created the Promise. The Runtime's runloop is always being pumped, so the block will execute.

This would require exposing `Runtime::RuntimeLoop()` to JS, perhaps via a global function:
https://github.com/NativeScript/ios/blob/73332e13667730b3084aa3dd7973685196efa086/NativeScript/runtime/Runtime.h#L27

adrian-niculescu:
A promise constructed while native code runs JS on a background thread never resolves. The proxy captures the current run loop at construction, and when the promise later resolves on a different thread it marshals the resolution back to that captured run loop. For a promise created on a background thread, that run loop is dormant once the call returns, so a resolution arriving on the runtime loop (for example from a `setTimeout`) is scheduled onto a loop nobody pumps and never runs.

This gates the marshaling on origin: a resolution is only marshaled back when the promise was created on the runtime loop, which is always being pumped. A promise created on any other thread resolves inline on whichever thread settles it. The same gate is applied to the get trap on the returned promise. This matches the approach discussed in the issue.

Test: `TestRunner/app/tests/Promises.js` gets a case that constructs a promise on a background dispatch queue and resolves it on the runtime loop. It times out before the change and passes after; the rest of the suite stays green.

Fixes #330

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Promise settlement and callback timing when promises are created on background threads and resolved on the main runtime loop.
  * Enhanced reliability so `.then` handlers run immediately when appropriate, or dispatch correctly when crossing the thread boundary.
* **Tests**
  * Added coverage for Promise resolution from a background queue and completion on the main `NSOperationQueue`.

### DIFF
--- a/NativeScript/runtime/PromiseProxy.cpp
+++ b/NativeScript/runtime/PromiseProxy.cpp
@@ -1,19 +1,36 @@
 #include "PromiseProxy.h"
+
+#include <CoreFoundation/CoreFoundation.h>
+
 #include "Helpers.h"
+#include "Runtime.h"
 
 using namespace v8;
 
 namespace tns {
 
+// Reports whether the calling thread runs the isolate's runtime loop. That loop
+// is where timers fire and is always being pumped, so a promise resolution is
+// marshaled back to its creating thread only when that thread is the runtime
+// loop; a promise created elsewhere settles on whichever thread resolves it.
+static void IsRuntimeRunloopCallback(const FunctionCallbackInfo<Value>& args) {
+    Runtime* runtime = Runtime::GetRuntime(args.GetIsolate());
+    bool isRuntimeLoop = runtime != nullptr && CFRunLoopGetCurrent() == runtime->RuntimeLoop();
+    args.GetReturnValue().Set(isRuntimeLoop);
+}
+
 void PromiseProxy::Init(v8::Local<v8::Context> context) {
     std::string source = R"(
-        // Ensure that Promise callbacks are executed on the
-        // same thread on which they were created
-        (() => {
+        // Run a Promise's callbacks on the thread that created it, but only when
+        // that thread is the runtime loop. A Promise created on a background
+        // thread settles on whichever thread resolves it, because the background
+        // run loop may be dormant and marshaling a resolution to it would hang.
+        (function(isRuntimeRunloop) {
             global.Promise = new Proxy(global.Promise, {
                 construct: function(target, args) {
                     let origFunc = args[0];
                     let runloop = CFRunLoopGetCurrent();
+                    let originIsRuntimeLoop = isRuntimeRunloop();
 
                     let promise = new target(function(resolve, reject) {
                         function isFulfilled() {
@@ -29,7 +46,7 @@ void PromiseProxy::Init(v8::Local<v8::Context> context) {
                                 return;
                             }
                             const resolveCall = resolve.bind(this, value);
-                            if (runloop === CFRunLoopGetCurrent()) {
+                            if (!originIsRuntimeLoop || runloop === CFRunLoopGetCurrent()) {
                                 markFulfilled();
                                 resolveCall();
                             } else {
@@ -42,7 +59,7 @@ void PromiseProxy::Init(v8::Local<v8::Context> context) {
                                 return;
                             }
                             const rejectCall = reject.bind(this, reason);
-                            if (runloop === CFRunLoopGetCurrent()) {
+                            if (!originIsRuntimeLoop || runloop === CFRunLoopGetCurrent()) {
                                 markFulfilled();
                                 rejectCall();
                             } else {
@@ -60,7 +77,7 @@ void PromiseProxy::Init(v8::Local<v8::Context> context) {
                                 return orig.bind(target);
                             }
                             return typeof orig === 'function' ? function(x) {
-                                if (runloop === CFRunLoopGetCurrent()) {
+                                if (!originIsRuntimeLoop || runloop === CFRunLoopGetCurrent()) {
                                     orig.bind(target, x)();
                                     return target;
                                 }
@@ -72,7 +89,7 @@ void PromiseProxy::Init(v8::Local<v8::Context> context) {
                     });
                 }
             });
-        })();
+        })
     )";
 
     Isolate* isolate = context->GetIsolate();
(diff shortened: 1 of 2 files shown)


adrian-niculescu:
> 🧹 Nitpick comments (1)
> > TestRunner/app/tests/Promises.js (1)> `142-147`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Fail explicitly on unexpected rejection.**
> > > If this promise rejects, the test currently hangs until timeout, which obscures the failure. Add a `catch` that fails and completes the async spec.
> > > 
> > > Proposed test hardening
> > > ```diff
> > >              new Promise(resolve => {
> > >                  NSOperationQueue.mainQueue.addOperationWithBlock(() => resolve("settled"));
> > >              }).then(value => {
> > >                  expect(value).toBe("settled");
> > >                  done();
> > > +            }).catch(error => {
> > > +                expect(true).toBe(false, `Promise rejected unexpectedly: ${error}`);
> > > +                done();
> > >              });
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@TestRunner/app/tests/Promises.js` around lines 142 - 147, The Promise test in
> > > the async spec currently only handles the success path, so unexpected rejection
> > > can hang the test until timeout. Update the Promise chain in the `Promises.js`
> > > test case to add a `catch` alongside the existing `then` so any rejection
> > > explicitly fails the spec and calls `done()`. Keep the change focused on the
> > > affected promise block near the
> > > `NSOperationQueue.mainQueue.addOperationWithBlock` usage and the
> > > `expect(value).toBe("settled")` assertion.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Nitpick comments:
> In `@TestRunner/app/tests/Promises.js`:
> - Around line 142-147: The Promise test in the async spec currently only handles
> the success path, so unexpected rejection can hang the test until timeout.
> Update the Promise chain in the `Promises.js` test case to add a `catch`
> alongside the existing `then` so any rejection explicitly fails the spec and
> calls `done()`. Keep the change focused on the affected promise block near the
> `NSOperationQueue.mainQueue.addOperationWithBlock` usage and the
> `expect(value).toBe("settled")` assertion.
> ```
> 
> ℹ️ Review info

@coderabbitai, fixed now.

adrian-niculescu:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> `@adrian-niculescu` Thanks for the update — marking this as addressed.
> 
> (ᵔᴥᵔ)
> 
> If you found this review helpful, would you consider giving us a shout-out on [X](https://twitter.com/intent/tweet?related=CodeRabbitAI&text=Great%20catch%20by%20%40coderabbitai%3A%20add%20a%20catch%20to%20fail%20fast%20on%20unexpected%20Promise%20rejections%2C%20so%20async%20tests%20don%E2%80%99t%20hang%20until%20timeout.%20Small%20fix%2C%20big%20reliability%20boost%20%F0%9F%A7%AA%E2%9A%A1%20Try%20it!&url=https%3A%2F%2Fgithub.com%2FNativeScript%2Fios%2Fpull%2F396)?
> 
> Thank you for using CodeRabbit!

@coderabbitai, please approve the PR if there's nothing else to address.

edusperoni (approved):

adrian-niculescu:
Thank you!

PR #403, opened 2026-07-09
feat: support native ES classes with lazy registration, including static usage before construction

NathanWalker:
Makes plain ES2015+ classes that extend native types work directly on iOS, without requiring `@NativeClass` or ES5 downleveling:

```ts
class JSClass extends NSObject {
  description() {
    return 'hello from ES class';
  }
}
```

## Lazy registration

Construction is not the only way a class first crosses into native code. All of the following now trigger lazy registration, before any instance has ever been created:

```ts
class JSClass extends NSObject {}

someNativeMethod(JSClass);   // passed as a Class (or id) argument
JSClass.alloc().init();      // alloc before any construction
JSClass.new();               // inherited static factory
JSClass.someStaticMethod();  // inherited static method dispatch
JSClass.someStaticProperty;  // inherited static property get/set
```

## Instance identity Improvements

`new MyClass()` and native `[[MyClass.class() alloc] init…]` now produce the same kind of JS instance: a real construct of the ES class. Public fields, private fields (`#a`), and the constructor body run on both paths.

The constructor → `super()` loop is broken with an isolate-local adopt slot (`Caches::PendingESAdopt`):

1. Native already created object **N1**.
2. `CreateJsWrapper(N1)` stashes N1 and constructs `MyClass`.
3. `super()` binds `this` to N1 and does **not** `alloc` / `init` again.

`super({ argName: value })` stays the create-path initializer picker. Adopt ignores those args so the initializer that already ran (`init`, `initWithFrame:`, `initWithCoder:`) stays authoritative.

```ts
class MyClass extends UIView {
  #a = 1;
  constructor() {
    super({ frame: CGRectMake(0, 0) });
  }
  someMethod() {
    return this.#a;
  }
}

new MyClass().someMethod();          // 1
MyClass.alloc().init().someMethod(); // 1  (was TypeError: cannot read private member #a)
```

Not solved here (not deal-breakers):

- Constructors that require JS-only arguments cannot be invented on the native-born path (`CreateJsWrapper` constructs with zero args).
- JS-only methods still need `ObjCExposedMethods` / `NativeClass({ ios: { methods } })` to be callable via `objc_msgSend`.
- A constructor that throws after `super()` leaves a partial `Instances` mapping; a later wrap of the same `id` does not construct again.
- KVO / isa-swizzled first wrap still falls back to the pre-A empty wrapper.

## NativeClass decorator API

`@NativeClass` is no longer a no-op. Optional `ios` options map to the existing statics and can eagerly name the Objective-C class:

```ts
@NativeClass({
  ios: {
    name: 'MyNeatIOSClass',
    protocols: [UIDelegateAnything],
    methods: {
      'selectorWithX:andY:': {
        returns: interop.types.void,
        params: [interop.types.id, interop.types.id],
      },
    },
  },
  // here for example purposes - consolidated decorator for all platform annotations 
  android: {
    interfaces: [android.view.View],
    name: 'org.nativescript.example.CustomActivity',
  },
})
class MyClass extends UIView {}
```

- All properties are optional.
- `ios.protocols` → `static ObjCProtocols`
- `ios.methods` → `static ObjCExposedMethods`
- `ios.name` → `ObjCClassName` plus eager registration
- This runtime implements `ios` only; `android` is accepted and ignored
- Types: [`types/ns-nativeclass.d.ts`](https://github.com/NativeScript/ios/blob/feat/native-es-classes/types/ns-nativeclass.d.ts)

`@NativeClass` and `@NativeClass({ ios: { … } })` are both valid. Passing the class directly (`NativeClass(MyClass)`) applies empty options.

## Tests

See [`TestRunner/app/tests/Inheritance/ESClassTests.js`](https://github.com/NativeScript/ios/blob/feat/native-es-classes/TestRunner/app/tests/Inheritance/ESClassTests.js), including:

- `ESClassAllocInitRunsJsConstructor`
- `ESClassAllocInitPrivateFields`
- `ESClassAllocInitDoesNotDoubleAlloc`
- `ESClassSuperObjectTokensSelectInitializer`
- `ESClassAllocInitThrowingConstructor`
- `NativeClassDecoratorAppliesIOSOptions`
- `NativeClassEagerNameRegistersImmediately`
- `NativeClassExposedMethodsFromIOSOptions`

## Summary by CodeRabbit

* **New Features**
  * Added `NativeClass` support for decorating JavaScript classes with native class names, protocols, interfaces, and exposed methods.
  * Added TypeScript declarations and configuration options for `NativeClass` on iOS and Android.
  * Improved interoperability between JavaScript subclasses and native classes, including inheritance, constructors, static members, properties, allocation, and class marshalling.
* **Bug Fixes**
  * Improved handling of `super()` construction and native object adoption for derived classes.
* **Tests**
  * Added comprehensive coverage for ES class inheritance and native integration.

### DIFF
--- a/TestRunner/app/tests/Inheritance/ESClassTests.js
+++ b/TestRunner/app/tests/Inheritance/ESClassTests.js
@@ -1,0 +1,287 @@
+// Tests for native ES class support: plain `class X extends NativeType {}` without the
+// @NativeClass decorator or ES5 downleveling. The Objective-C class is registered lazily by
+// the runtime on first use (construction, alloc/new, static dispatch or Class marshalling).
+describe(module.id, function () {
+    afterEach(function () {
+        TNSClearOutput();
+    });
+
+    it('ESClassLazyRegistration', function () {
+        class ESLazyObject extends NSObject {
+        }
+
+        // Defining the class must not register anything with the Objective-C runtime
+        expect(NSClassFromString('ESLazyObject')).toBeNull();
+
+        var instance = new ESLazyObject();
+        expect(instance instanceof ESLazyObject).toBe(true);
+
+        // First construction registers the class under the ES class name
+        expect(NSClassFromString('ESLazyObject')).toBe(ESLazyObject);
+    });
+
+    it('ESClassSimpleInheritance', function () {
+        class ESSimpleObject extends TNSDerivedInterface {
+        }
+
+        var object = new ESSimpleObject();
+        expect(object.constructor).toBe(ESSimpleObject);
+        expect(object instanceof ESSimpleObject).toBe(true);
+        expect(object instanceof TNSDerivedInterface).toBe(true);
+        expect(object instanceof NSObject).toBe(true);
+        expect(object.class()).toBe(ESSimpleObject);
+        expect(object.superclass).toBe(TNSDerivedInterface);
+        expect(ESSimpleObject.class()).toBe(ESSimpleObject);
+        expect(ESSimpleObject.superclass()).toBe(TNSDerivedInterface);
+        expect(NSStringFromClass(ESSimpleObject)).toBe('ESSimpleObject');
+    });
+
+    it('ESClassConstructorLogicAndFields', function () {
+        class ESConstructorObject extends NSObject {
+            field = 42;
+
+            constructor() {
+                super();
+                this.initialized = true;
+            }
+        }
+
+        var object = new ESConstructorObject();
+        // The receiver created by super() must be the native-backed instance, with class
+        // fields and constructor logic applied to it
+        expect(object.field).toBe(42);
+        expect(object.initialized).toBe(true);
+        expect(object instanceof ESConstructorObject).toBe(true);
+        expect(object instanceof NSObject).toBe(true);
+        expect(NSStringFromClass(object.class())).toBe('ESConstructorObject');
+    });
+
+    it('ESClassInstanceMethodsAndSuper', function () {
+        class ESMethodsObject extends TNSDerivedInterface {
+            baseMethod() {
+                TNSLog('js baseMethod called');
+                super.baseMethod();
+            }
+            derivedMethod() {
+                TNSLog('js derivedMethod called');
+                super.derivedMethod();
+            }
+        }
+
+        var object = new ESMethodsObject();
+        object.baseMethod();
+        object.derivedMethod();
+        expect(TNSGetOutput()).toBe('js baseMethod called' +
+            'instance baseMethod called' +
+            'js derivedMethod called' +
+            'instance derivedMethod called');
+    });
+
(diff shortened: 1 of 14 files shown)


TestRunner/app/tests/Inheritance/ESClassTests.js
```diff
@@ -0,0 +1,287 @@
...
+
+    it('ESClassSimpleInheritance', function () {
+        class ESSimpleObject extends TNSDerivedInterface {
+        }
+
+        var object = new ESSimpleObject();
+        expect(object.constructor).toBe(ESSimpleObject);
+        expect(object instanceof ESSimpleObject).toBe(true);
+        expect(object instanceof TNSDerivedInterface).toBe(true);
+        expect(object instanceof NSObject).toBe(true);
```
edusperoni, line 31: does this call `alloc().init()`? I'm not sure about the implications of calling `new Something()` when the native object might have different ideas for initalization.

would `ESSimpleObject.alloc().init()` call the constructor? what about the constructor arguments?

NathanWalker: Yes in the default case. super() lands in the exact same ArgConverter::ConstructObject path that every NativeScript construction (new NSObject(), legacy .extend() classes) has always used. The only new behavior is that new.target resolution makes it alloc the derived ObjC class instead of the base. The dispatch inside ConstructObject is:
```
if (result == nil && interfaceMeta != nullptr && info.Length() > 0) {
  std::vector<Local<Value>> args;
  const MethodMeta* initializer =
      ArgConverter::FindInitializer(context, klass, interfaceMeta, info, args);
  result = [klass alloc];

  V8VectorArgs vectorArgs(args);
  result = Interop::CallInitializer(context, initializer, result, klass, vectorArgs);
}

if (result == nil) {
  result = [[klass alloc] init];
}
```

So for a native class with "different ideas about initialization," the existing constructor-to-initializer matching still applies: super() with no arguments is literally [[DerivedClass alloc] init], while super(args...) runs FindInitializer, which matches the arguments against the class's initWith… selectors from metadata and calls the matched designated initializer. The key detail is that what reaches the native initializer is whatever you pass to super(...), not what's passed to new. The JS constructor body is in control, same as any ES class. If a native base has no usable zero-arg init, the author passes matching args to super(...) or uses the explicit alloc().initWithX(...) pattern, exactly as before.

`alloc()` triggers lazy registration (creating the ObjC class and installing method/accessor overrides; a metadata operation only) and returns an uninitialized instance; .init() is then an ordinary message send.

NathanWalker: Pushed up 2 additional test cases that hopefully clarify that: https://github.com/NativeScript/ios/pull/403/commits/75a9934b158ce842b4d04bb568b9382ca25ab122

PR #408, opened 2026-07-21
feat: update libffi to upstream with FFI_TYPE_VECTOR support

edusperoni:
## What

Bumps the `libffi` submodule from the 2019-era fork (`darind/v8-ios`, 3.3-rc0 + EXT_VECTOR patch) to [`feat/vector-types`](https://github.com/NativeScript/libffi/tree/feat/vector-types): **current upstream libffi** plus the `FFI_TYPE_VECTOR` patch series proposed upstream (supersedes libffi/libffi#414, fixes the aarch64 vector marshaling tracked in libffi/libffi#773).

## Why

- **ObjC exceptions can now unwind through ffi closure trampolines** into a native `@try/@catch`. With the old fork this aborted the process (missing unwind coverage for the `FFI_EXEC_TRAMPOLINE_TABLE` return path). This unblocks forwarding JS-originated exceptions to native callers (planned `interop.escapeException`).
- **Six years of upstream libffi fixes** (the fork had been frozen since 2019).
- **Vector (SIMD) marshaling moves to the upstream API**: `FFI_TYPE_VECTOR` (now 18 — upstream claimed 16/17 for `UINT128`/`SINT128`) described by lane count via `elements[]`, with libffi computing the ABI classification. `FFICall.cpp` prefills the storage layout (pow2-rounded size, `min(size,16)` alignment — exactly what libffi derives) because the runtime reads `ffi_type.size` outside prepared cifs; `ffi_prep_cif` honors prefilled aggregates. The manual 3→4 lane padding hack and the legacy `FFI_TYPE_EXT_VECTOR` path are gone.
- **`MatrixTests` re-enabled**: the historical `EXC_BAD_ACCESS` in matrix marshaling no longer reproduces with the new homogeneous-vector-aggregate classification.

## build_libffi.sh rewrite

- Uses the selected Xcode via `xcrun`/`xcode-select` (respects `DEVELOPER_DIR`) instead of hardcoded `/Applications/Xcode.app` paths.
- Builds **all six shipped slices**, now including `arm64-xros` and `arm64-xrsimulator`.
- Modern autotools layout (out-of-tree builds, `.libs/libffi.a`), `--disable-docs`, no `-fembed-bitcode`, iOS 13.0 / visionOS 2.0 minimums.
- `--install` copies artifacts into `NativeScript/lib/<slice>/` and `NativeScript/include/libffi/<arch>/`.

All vendored slices and headers in this PR were produced by the script.

## Known limitation

On x86_64 (`iphonesimulator`, `maccatalyst`) vectors wider than 16 bytes are rejected with `FFI_BAD_TYPEDEF` instead of marshaled — correct `%ymm` passing needs `unix64.S` register-save work upstream. Accepted as a sunsetting-architecture tradeoff; arm64 targets are unaffected (up to 32-byte vectors + HVAs fully supported).

## Validation

- **iOS simulator: 869 tests, 0 failures**, including `Marshalling/VectorTests` (10/10) and the re-enabled `Marshalling/MatrixTests` (20/20).
- **visionOS 26.2 simulator: 869 tests, 0 failures** with the same libffi content (run together with the visionOS test enablement from #407).
- libffi's own new `libffi.vector` suite: 40/40 on both arm64 and x86_64; existing `libffi.call` suite regression-free.
- Standalone unwinding tests: ObjC exceptions thrown through `ffi_call` **and** from closure callbacks through the trampoline are catchable natively on arm64 (both directions previously impossible for closures).

## Summary by CodeRabbit

* **New Features**
  * Added reusable call-plan APIs for preparing and invoking native calls.
  * Expanded support for vector types, 128-bit integers, and complex numeric types.
  * Updated native libraries across supported Apple platforms.

* **Improvements**
  * Improved ARM64 and x86_64 ABI and platform compatibility.
  * Added configurable multi-platform library builds with parallel job support.

* **Deprecations**
  * Marked legacy Java raw-ABI helper APIs as deprecated.

* **Tests**
  * Re-enabled matrix marshalling tests.


edusperoni:
Addressed both review findings in 2aed0b3:

- **`ffitarget.h` include**: restored the quoted include (`#include "ffitarget.h"`) that the pre-update vendored headers used — the upstream-generated angle include only resolved through Xcode's basename headermaps, which is ambiguous with both arch copies of `ffitarget.h` in the project (a silent cross-arch include hazard for x86_64 builds). `build_libffi.sh --install` now applies the same rewrite automatically so future header regenerations keep it.
- **`--jobs` validation**: missing or non-numeric values are rejected with a usage error instead of tripping `set -u` or producing a broken `make -j`.

NathanWalker (approved):

PR #413, opened 2026-07-27
feat: Ada v4

NathanWalker:
- Updates to latest Ada v4

https://yagiz.co/release-of-ada-v4/


anonrig (approved):

anonrig:
We expose a max url length method now. If you want to have smaller URL size to accept, you can now use it.

PR #415, opened 2026-07-30
feat: Node-style primordials for runtime builtins

edusperoni:
Stacked on #411 — review that one first; this PR targets `feat/js-builtins`.

## What this adds

The runtime's builtin JavaScript installs globals and leaves closures behind that run for the lifetime of the app: event dispatch, the `Promise` proxy traps, console's smart-stringify, `__extends`, the `URL.searchParams` accessor. Until now those closures reached for intrinsics (`Array.prototype.slice`, `JSON.stringify`, `Object.create`, …) through the live globals, so app code replacing one could break runtime internals or observe them.

`NativeScript/runtime/js/primordials.js` is a new builtin that captures exactly the intrinsics the other builtins need into a frozen, null-prototype namespace, Node-style.

## The `(binding, primordials)` contract

Builtins are now compiled as function bodies with **two** fixed parameters:

```js
const { someNative } = binding;
const { ArrayPrototypeSlice, ObjectCreate } = primordials;
```

`primordials` is built lazily on the first `RunBuiltin` of an isolate — which happens during runtime init, before any user code — and cached in `Caches::Primordials`. Consequences:

- Worker isolates snapshot their own realm's intrinsics automatically.
- Builtins that compile late still get the pristine snapshot. `smart-stringify` is only compiled on the first object logged, which can be well after user code has run.
- `binding` is unchanged; most builtins still receive `undefined` for it.
- The bytecode cache stays sound: it is process-local and in-memory only, and every compile in a process now uses the same 2-parameter signature.

## Why uncurrying, uniformly

Instance methods are exposed uncurried, exactly as Node does it:

```js
const uncurryThis = Function.prototype.bind.bind(Function.prototype.call);
// ArrayPrototypeSlice(list, 1)          not  list.slice(1)
// FunctionPrototypeCall(cb, this, ev)   not  cb.call(this, ev)
```

The runtime always runs V8 **jitless** on device, so the usual "uncurrying is free once optimized" argument does not apply and the cost had to be measured. Benchmarked on Node 24 / V8 13.6, arm64 host, `--jitless`:

- Uncurried primordial call: **+5-12% per op** versus a raw method call.
- For event dispatch that is ~+10%, about **26 ns on a 4-listener dispatch** — irrelevant next to the listener bodies.
- Under Ignition, uncurried **beats** the `captured.call(recv, …)` alternative, which is why uncurrying is used uniformly rather than mixing styles.
- Captured statics (`JSON.stringify` etc.) are free.
- `Reflect.apply` is measurably worse and is not used; `FunctionPrototypeApply` covers those sites.

Plain constructor calls made once at init time (`new Map()` while bootstrapping) may stay direct — the rule targets code in closures that outlive init.

## Enforcement

`eslint.config.mjs` declares `primordials` and adds two rules over `NativeScript/runtime/js`, both verified to fire:

- `no-restricted-properties` for every captured **static** (`JSON.stringify`, `Object.defineProperty/assign/freeze/create/keys/setPrototypeOf/getOwnPropertyDescriptor`, `Array.from`, `Reflect.construct`, `Reflect.apply`).
- `no-restricted-globals` for every captured **constructor** (`Error`, `Map`, `Proxy`, `TypeError`). A `const { Proxy } = primordials` destructure shadows the global, so the rule only fires on an unguarded reference.

Each message names the replacement. `primordials.js` itself is exempted via a per-file override. Uncurried instance-method use cannot be matched by receiver, so that stays a review rule — documented as such in `NativeScript/runtime/js/README.md`.

Deliberately left as live global lookups: `Reflect.decorate` in `inline-functions.js` (`reflect-metadata` installs it after runtime init, and the TypeScript `__decorate` shim must see it), `Blob`/`File` in `blob-url.js` (provided by the app layer, not the runtime), and `String(x)`/`instanceof Array` conversions where tamper-immunity is not affected.

`ts-helpers` builds native super calls with the captured `Reflect.construct` rather than `bind`-then-`new`, which keeps that path off the array iterator protocol as well.

## Tests

New `TestRunner/app/tests/PrimordialsTests.js` (7 specs) tampers with the intrinsics and checks the runtime keeps working. Each spec keeps the tampered window synchronous and assertion-free, restoring the originals in a `finally`:

- a guard spec proving the tampering is actually observable, so the suite cannot pass vacuously
- global `dispatchEvent` delivers to every listener (function and `handleEvent`) with `Array.prototype.slice/indexOf/push/splice` and `Function.prototype.call` replaced
- `addEventListener`/`removeEventListener` and `once` semantics under the same tampering
- `reportError` still reaches an `error` listener with a correct `ErrorEvent`
- `console.log` of a circular object stays non-fatal with `JSON.stringify` and the replacer's `Array.prototype.indexOf/push` replaced (the smart-stringify path; its output is not reachable from JS and `JsonStringifyObject` swallows a throwing stringify, so this spec is a crash/throw check rather than an output assertion — noted in the test)
- a native `_super.call(this, x)` over an Objective-C base with `Array.prototype.slice/concat`, `Function.prototype.bind` and `Reflect.construct` replaced
- `__extends` and `__tsEnum` with the `Object` statics replaced

**Result: 912 specs, 0 failures** (905 baseline + 7), independently reproduced on a second simulator. ESLint clean; every builtin verified to compile as a `(binding, primordials)` function body.

## Known follow-ups (not in this PR)

Pre-existing live-intrinsic reads that primordials does not yet cover: `for...of ObjectKeys(...)` in `__tsEnum` and `ArrayFrom(arguments)` in `inline-functions.js` still go through the array iterator protocol, and `String(x)` conversions in `events.js`/`error-events.js` read the live global `String`.

## Summary by CodeRabbit

* **New Features**
  * Added a frozen snapshot of JavaScript built-ins for runtime components, captured before application code can modify global behavior.
  * Runtime built-ins now consistently use protected intrinsic operations, improving reliability when global methods are overridden.

* **Bug Fixes**
  * Improved event handling, promises, inheritance, serialization, and helper behavior under modified JavaScript globals.

* **Tests**
  * Added coverage confirming runtime functionality remains stable when built-in methods are replaced.

### DIFF
--- a/NativeScript/runtime/BuiltinLoader.cpp
+++ b/NativeScript/runtime/BuiltinLoader.cpp
@@ -3,6 +3,7 @@
 #include <mutex>
 #include <vector>
 
+#include "Caches.h"
 #include "Helpers.h"
 
 using namespace v8;
@@ -17,12 +18,14 @@ std::vector<uint8_t> builtinCache[static_cast<unsigned>(BuiltinId::kCount)];
 
 // Every builtin is compiled as a function body receiving these fixed
 // parameters, mirroring Node's module wrapper: a file exports through
-// `module.exports`/`exports`, and natives arrive as properties of the
-// `binding` bag (Node's internalBinding idiom) for each file to destructure.
+// `module.exports`/`exports`, natives arrive as properties of the `binding`
+// bag (Node's internalBinding idiom) and intrinsics as properties of
+// `primordials`; each file destructures what it needs.
 constexpr const char* kExportsParamName = "exports";
 constexpr const char* kModuleParamName = "module";
 constexpr const char* kBindingParamName = "binding";
-constexpr int kParamCount = 3;
+constexpr const char* kPrimordialsParamName = "primordials";
+constexpr int kParamCount = 4;
 
 MaybeLocal<v8::Function> CompileBuiltin(Local<Context> context, BuiltinId id) {
   Isolate* isolate = v8::Isolate::GetCurrent();
@@ -49,9 +52,11 @@ MaybeLocal<v8::Function> CompileBuiltin(Local<Context> context, BuiltinId id) {
   );
   Local<v8::String> sourceText = tns::ToV8String(
       isolate, builtin.source, static_cast<int>(builtin.length));
-  Local<v8::String> params[] = {tns::ToV8String(isolate, kExportsParamName),
-                                tns::ToV8String(isolate, kModuleParamName),
-                                tns::ToV8String(isolate, kBindingParamName)};
+  Local<v8::String> params[] = {
+      tns::ToV8String(isolate, kExportsParamName),
+      tns::ToV8String(isolate, kModuleParamName),
+      tns::ToV8String(isolate, kBindingParamName),
+      tns::ToV8String(isolate, kPrimordialsParamName)};
 
   Local<v8::Function> fn;
   if (!blob.empty()) {
@@ -91,11 +96,8 @@ MaybeLocal<v8::Function> CompileBuiltin(Local<Context> context, BuiltinId id) {
   return fn;
 }
 
-}  // namespace
-
-MaybeLocal<Value> BuiltinLoader::RunBuiltin(Local<Context> context,
-                                            BuiltinId id,
-                                            Local<Value> binding) {
+MaybeLocal<Value> CallBuiltin(Local<Context> context, BuiltinId id,
+                              Local<Value> binding, Local<Value> primordials) {
   Isolate* isolate = v8::Isolate::GetCurrent();
 
   Local<v8::Function> fn;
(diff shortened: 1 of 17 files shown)


edusperoni:
Orchestrator review pass (folded into the commit):

- Independently verified the `Reflect.construct` switch in ts-helpers (bound-function newTarget semantics and the arguments-object concat edge both hold) and every other conversion.
- Closed the two flagged coverage gaps: `String` is now a captured primordial (used by the Event/ErrorEvent constructors and dispatch closures), and the two post-init paths that depended on the tamperable array-iterator protocol — `__tsEnum`'s `for...of` and the ObjC decorators' `Array.from(arguments)` — now use index loops. `Array.from` deliberately has no primordial; the ESLint config documents why and restricts `String` as a global.
- Suite after changes: **912/912**.

Deferred (pre-existing, unchanged by this PR): `child[Symbol.hasInstance] = ...` in ts-helpers is likely a silent no-op in sloppy mode (inherited non-writable `Function.prototype[@@hasInstance]`), and `instanceof Array` in ObjCClass could become `ArrayIsArray` — both worth their own look since fixing them changes behavior.

NathanWalker (approved):

PR #422, opened 2026-07-31
fix(url): resync cached searchParams when search/href changes

edusperoni:
Fixes two defects in `URL.prototype.searchParams` reported by CodeRabbit on #411: https://github.com/NativeScript/ios/pull/411#discussion_r3687911515

## The defects

1. **Staleness.** The getter cached a `URLSearchParams` built from the query string on first access. A later direct assignment to `url.search` or `url.href` did not refresh it, so subsequent `url.searchParams` reads returned parameters from the old query. Worse, mutating through a reference held across such an assignment wrote the stale set back onto the URL, silently discarding the newly assigned query.
2. **Enumeration leak.** The cache was stored as a plain enumerable own property, so `_searchParams` surfaced in `Object.keys(url)` and `JSON.stringify(url)`.

## Why resync in place instead of rebuilding

CodeRabbit's proposed diff rebuilds the cache into a **new** `URLSearchParams` when the recorded query no longer matches. That breaks the WHATWG same-object guarantee — `url.searchParams` must return the identical object for the URL's lifetime, and app code routinely holds a reference to it. A rebuild would leave every existing reference pointing at an object detached from its URL.

This PR keeps exactly one `URLSearchParams` per URL forever. When the recorded source string differs from the current `this.search`, the existing object is resynced in place: every key deleted, then re-appended from parsing the current query, then the recorded source updated. The resync goes through the raw methods captured at construction (`_append`, `_delete`, `_forEach`), never the public ones — those are wrapped to write back into `_url.search` and would clobber the very string being applied. The wrapped mutators also update the recorded source after writing back, so a mutation through the object never makes the next read look stale, and they resync first so a mutation on a reference held across a reassignment no longer clobbers the new query.

`_searchParams` and `_searchParamsSource` are defined non-enumerable via the destructured `ObjectDefineProperty`; all new intrinsic access goes through primordials (`ArrayPrototypePush`, `FunctionPrototypeCall`, both already captured — no additions to `primordials.js` or the eslint list were needed).

## Tests

New `TestRunner/app/tests/UrlSearchParamsTests.js` (12 specs), registered in `TestRunner/app/tests/index.js`:

- same-object identity across reads, and across a `url.search = ...` assignment
- refresh after `url.search` assignment, after `url.href` assignment, and when the query is cleared
- duplicate keys preserved across a refresh
- mutation write-back (`set`/`append`/`delete`/`sort` update `url.search` / `url.href`)
- mutation through the object does not trigger a resync that loses state on the next read
- mutating a held reference after a query reassignment does not clobber the new query
- `Object.keys(url)`, `JSON.stringify(url)` and `for..in` do not expose `_searchParams` / `_searchParamsSource`
- a fresh URL's first `searchParams` read parses correctly (including `getAll` and `size`)

## Drive-by: ErrorEvents spec hardening

Three specs in `ErrorEventsTests.js` asserted that **nothing at all** reached the process-wide `__onUncaughtError` hook during their quiet window. That is inherently order-dependent — an async failure from an earlier suite (the deliberate module-not-found in `NodeBuiltinsAndOptionalModulesTests.mjs`) can drain into the spy while the test waits, and the extra specs added here shifted timing enough to make it land there. They now assert that the specific error under test did not reach the hook, which is what they actually mean.

## Suite results

`./run_tests.sh` on the iOS simulator:

- baseline (`origin/main`): **922 tests, 0 failures**
- this branch: **934 tests, 0 failures**

ESLint (`npx eslint 'NativeScript/runtime/js/**/*.js'`) clean; `tools/js2c.mjs` regen clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `URL.searchParams` synchronization when URLs are updated through `search` or `href`.
  * Preserved stable parameter references while correctly reflecting query-string changes.
  * Ensured parameter mutations consistently update the associated URL.

* **Tests**
  * Added comprehensive coverage for caching, duplicate keys, empty queries, refresh behavior, and mutation synchronization.
  * Refined error-event tests to verify that prevented errors are excluded from uncaught-error reports.

### DIFF
--- a/NativeScript/runtime/js/blob-url.js
+++ b/NativeScript/runtime/js/blob-url.js
@@ -1,4 +1,6 @@
 const {
+    ArrayPrototypePush,
+    FunctionPrototypeCall,
     Map,
     MapPrototypeDelete,
     MapPrototypeGet,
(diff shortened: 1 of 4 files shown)


NathanWalker:
@coderabbitai could you investigate the failing `ApiTests` in the CI test run?

NathanWalker (approved):

PR #473, opened 2026-09-07
fix(worker): lifecycle fixes for workers that fail during startup

adrian-niculescu:
A worker that fails while starting can leave things behind in three ways.

A worker whose entry script does not exist reports the error but keeps running: the startup callback returns without terminating, so the looper enters its runloop and the thread and its isolate stay alive until the parent calls `terminate()`. A parent that only logs the error keeps the worker forever. The worker now terminates itself right after reporting, the way a worker whose entry throws already does.

The worker's registry entry is inserted only after `Start()` has queued the startup operation, because the wrapper takes its id inside `Start()`. A worker whose entry calls `close()` or exhausts its heap right away can delete its runtime before the entry exists: `~Runtime` dereferences the lookup unchecked, and the constructor then publishes an entry for a wrapper that is already disposed. The id is now taken in the wrapper's constructor and the entry is inserted before `Start()`. `isRunning_` is set before the operation is queued for the same reason, so a worker that already terminated is not flipped back to running, and `~Runtime` tolerates an absent entry.

Errors forwarded to `worker.onerror` are posted to the parent's loop as a task that captures the wrapper and reads its main isolate and persistent when it runs. A worker whose entry throws reports, terminates and is disposed on its own thread, after which the parent's finalizer deletes the wrapper the next time the Worker object is collected, so a worker nobody holds a reference to can lose its wrapper before the report runs. The task now captures the isolate and the persistent by value and returns when the handle is empty. The `TryCatch` overload carried a copy of the same task and shares the one path.

No new tests: the first case is covered by the existing missing-script test, where the `terminate()` in the handler becomes a no-op, and the other two depend on thread and GC timing. TestRunner passes on an iPhone 17 simulator.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worker startup and shutdown reliability.
  * Prevented crashes when worker state is unavailable during teardown.
  * Worker scripts that cannot be found now report an error and terminate correctly instead of remaining active.
  * Improved error handling for asynchronous worker operations, including cases where a worker has already been destroyed.

### DIFF
--- a/NativeScript/runtime/DataWrapper.h
+++ b/NativeScript/runtime/DataWrapper.h
@@ -652,11 +652,9 @@ class WorkerWrapper : public BaseDataWrapper {
                                  const std::string& source,
                                  const std::string& stackTrace, int lineNumber,
                                  bool async);
-  v8::Local<v8::Object> ConstructErrorObject(v8::Local<v8::Context> context,
-                                             std::string message,
-                                             std::string source,
-                                             std::string stackTrace,
-                                             int lineNumber);
+  static v8::Local<v8::Object> ConstructErrorObject(
+      v8::Local<v8::Context> context, std::string message, std::string source,
+      std::string stackTrace, int lineNumber);
 };
 
 }  // namespace tns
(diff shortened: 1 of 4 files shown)


edusperoni (approved):
Thank you!

